PR #1, opened 2019-02-20
fix: tooltip disappearance and stickiness

xtinec:
🐛 Bug Fix
This is a replication of PR #6898 in apache/incubator-superset

Nvd3 attaches tooltips to the body of the dom, not the chart the tooltip is meant fo. On hover, it sets their opacity to 1. In order to address both their stickiness when chart reloads (PR #6805) and thier disappearance on scroll in dashboards (PR #6852), we introduce a shouldRemove parameter to `hideTooltips` and only remove them befor chart reloads. For the scroll events triggered on dashboards, we only hide the tooltips by setting their opacity to 0. When they get hovered over again, nvd3 sets their opacity to 1 which causes them to reappear.

### DIFF
--- a/packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
+++ b/packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
@@ -737,8 +737,9 @@ function nvd3Vis(element, props) {
         .attr('height', height)
         .call(chart);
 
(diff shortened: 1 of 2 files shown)


williaster (approved):
LGTM ~I think krist is working on a build fix / you might need to rebase after that~

EDIT: looks like you should remove the generator approach and simply iterate over the targets to fix the build.

PR #4, opened 2019-02-25
build: add commitlint and commitizen

kristw:
### Internal
* Add `commitlint` hook for linting git commits locally and add `commitizen` command line tools for helping create commit message (Use this instead of `git commit`)
### Documentation
* Add section about commit message to readme

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,0 +1,1 @@
+module.exports = {extends: ['@commitlint/config-conventional']};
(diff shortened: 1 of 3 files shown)


commitlint.config.js
```diff
@@ -0,0 +1 @@
+module.exports = {extends: ['@commitlint/config-conventional']};
```
williaster: nit -- `{...}` => with spaces `{ ... }`

williaster (approved):

PR #7, opened 2019-03-04
fix(legacy-preset-chart-nvd3): redraw bar values after legend change

betodealmeida:
🐛 Bug Fix

Similar to https://github.com/apache-superset/superset-ui-plugins/pull/6, bar values are never redrawn. Because of that, interactions with the legend are broken.

### DIFF
--- a/packages/superset-ui-legacy-preset-chart-nvd3/src/utils.js
+++ b/packages/superset-ui-legacy-preset-chart-nvd3/src/utils.js
@@ -27,6 +27,8 @@ import { smartDateFormatter } from '@superset-ui/time-format';
 // (1 hour offset, 2 days offset, etc.)
 const TIME_SHIFT_PATTERN = /\d+ \w+ offset/;
 
+const ANIMATION_TIME = 1000;
+
 export function cleanColorInput(value) {
   // for superset series that should have the same color
   return String(value)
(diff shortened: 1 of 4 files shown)


packages/superset-ui-legacy-preset-chart-nvd3/src/utils.js
```diff
@@ -58,29 +60,31 @@ export function drawBarValues(svg, data, stacked, axisFormat) {
...
-          .text(format(stacked ? totalStackedValues[index] : d.y))
-          .attr('transform', transformAttr)
-          .attr('class', 'bar-chart-label');
-        const labelWidth = textEls.node().getBBox().width;
-        textEls.attr('x', xPos + rectWidth / 2 - labelWidth / 2); // fine tune
-      }
-    });
+  svg.selectAll('.bar-chart-label').remove();
+  setTimeout(() => {
+    const groupLabels = svg.select('g.nv-barsWrap').append('g');
```
kristw: This will create new `g` every time. Is that expected?

betodealmeida: Ah, good point. I'll fix this.

packages/superset-ui-legacy-preset-chart-nvd3/src/utils.js
```diff
@@ -58,29 +60,34 @@ export function drawBarValues(svg, data, stacked, axisFormat) {
...
-          .attr('class', 'bar-chart-label');
-        const labelWidth = textEls.node().getBBox().width;
-        textEls.attr('x', xPos + rectWidth / 2 - labelWidth / 2); // fine tune
-      }
-    });
+  svg.selectAll('.bar-chart-label').remove();
+  setTimeout(() => {
+    const groupLabels = svg
+      .select('g.nv-barsWrap')
+      .append('g')
```
kristw, line 67: This line still adds new `g`

betodealmeida: Sorry, I don't know enough about SVG to know if this is bad. Is the problem that we're adding a `g`, or that every call will add **another** `g`? Because since it has the `'bar-chart-label` class, it will get removed next time (see line 63), so we'll add at most one `g`.

kristw: I took a closer look. This is ok, but let's use another `class` name that is not `bar-chart-label` e.g. `bar-chart-label-group` because this will be the same with the `text` in `.each()` below

betodealmeida: Cool, let me fix this. Thanks, @kristw!

kristw (approved):

PR #9, opened 2019-03-05
chore: lint the webpack config file in storybook

kristw:
🏠 Internal

Enable hot reload for typescript in storybook so you can edit src and see changes immediately


xtinec (commented):
hmm, only seeing lint changes. More changes needed to enable hot reload, perhaps?

kristw:
> hmm, only seeing lint changes. More changes needed to enable hot reload, perhaps?

Ah. Just realize I have already enabled this previously and this was only fixing lint.

williaster (approved):
lgtm if you still want to merge it

PR #11, opened 2019-03-11
[WIP] new line chart

kristw:
> This PR is being split into smaller pieces for ease of reviewing. 

🏆 Enhancements

### Define encoding config for a visualization

* Each visualization has one or more *channels* that can be encoded, such as `x`, `y`, `color`, `stroke`, fill. These channels configurations should be consistent/sharable across all visualizations. For example, `x` and `y` usually have scales and axes. Scales have `domain`, `range`. Axes have `label`, `format`. Color can support categorical, sequential or fixed color. All channels can support fixed value, field name in the data, or sometimes aggregates (haven't implemented the aggregate yet).
* Each encoding channel config is based on `vega-lite` encoding grammar. ([See line chart example](https://github.com/apache-superset/superset-ui-plugins/compare/kristw--line?expand=1#diff-89f0892b8e6df5ff19a42243e2254d6a), [vega-lite encoding documentation](https://vega.github.io/vega-lite/docs/encoding.html))
* Only adopt a small subset of `vega-lite` encoding grammar for now as we don’t have enough time (and may not intend) to provide 100% feature of `vega-lite`. 

### Implement TypeScript types for the encoding config.

* First attempt was using all types from `vega-lite` package (`XXXFieldDef`) directly but ended up getting too overwhelmed.
* Now adopting the primitive types from `vega-lite` then write higher-level types that has same name but only with the subset of the fields, so we can keep that of what we support and slowly adopt more in the future instead of taking it all now which our encoder does not know how to handle and also make it harder to implement the encoder. ([preview](https://github.com/apache-superset/superset-ui-plugins/compare/kristw--line?expand=1#diff-14e0823f4ab7530efbab63a9183f6d51))

### Write an encoder that parses the encoding config mention above and become a helper during visualization rendering.

* There is a base `Encoder` class and each visualization will implement its own `Encoder` that extends from the base class. See: [base Encoder class](https://github.com/apache-superset/superset-ui-plugins/compare/kristw--line?expand=1#diff-16f12a4b98393d4c5370f6b564777c0b), [Line Chart's Encoder](https://github.com/apache-superset/superset-ui-plugins/compare/kristw--line?expand=1#diff-73daa5efbd07552f1824a8c865377070)
* End up rewriting a few times to get cleaner APIs.  
* Still not fully satisfied and will do another pass this week.

### Long shot

* Once all these encoding configs code are a bit more stable, it could be extracted to `@superset-ui/encodeable`.


kristw:
Split into multiple PRs and reborn as #26

PR #12, opened 2019-03-12
 feat: add typescript declaration for external packages

kristw:
🏠 Internal

add declaration for external packages `@data-ui` and `@vx`

### DIFF
--- a/packages/superset-ui-preset-chart-xy/types/@data-ui/theme/index.d.ts
+++ b/packages/superset-ui-preset-chart-xy/types/@data-ui/theme/index.d.ts
@@ -1,0 +1,110 @@
+declare module '@data-ui/theme' {
+  type SvgLabelTextStyle = {
+    fontFamily: string;
+    fontSize: number;
+    fontWeight: number;
+    letterSpacing: number;
+    fill: string;
+    stroke: string;
+    textAnchor?:
+      | '-moz-initial'
+      | 'inherit'
+      | 'initial'
+      | 'revert'
+      | 'unset'
+      | 'end'
+      | 'start'
+      | 'middle';
+    pointerEvents?:
+      | '-moz-initial'
+      | 'inherit'
+      | 'initial'
+      | 'revert'
+      | 'unset'
+      | 'auto'
+      | 'none'
+      | 'visible'
+      | 'all'
+      | 'fill'
+      | 'stroke'
+      | 'painted'
+      | 'visibleFill'
+      | 'visiblePainted'
+      | 'visibleStroke';
+  };
+
+  export interface ChartTheme {
+    colors: {
+      default: string;
+      dark: string;
+      light: string;
+      disabled: string;
+      lightDisabled: string;
+      text: string;
+      black: string;
+      darkGray: string;
+      lightGray: string;
+      grid: string;
+      gridDark: string;
+      label: string;
+      tickLabel: string;
+      grays: string[];
+      categories: string[];
+    };
+    labelStyles: SvgLabelTextStyle & {
+      color: string;
+      lineHeight: string;
+      paddingBottom: number;
+      paddingTop: number;
+    };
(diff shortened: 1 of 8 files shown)


packages/superset-ui-preset-chart-xy/types/@data-ui/theme/index.d.ts
```diff
@@ -0,0 +1,110 @@
...
+        bottom: SvgLabelTextStyle;
+        top: SvgLabelTextStyle;
+      };
+    };
+    xTickStyles: {
+      stroke: string;
+      length: number;
+      label: {
+        bottom: SvgLabelTextStyle & {
+          dy: string;
```
williaster, line 77: don't want to get too nit-picky on this / not blocking: curious why `dx`/`dy` couldn't be added to `SvgLabelTextStyle ` as optional?

packages/superset-ui-preset-chart-xy/types/@vx/legend/index.d.ts
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable react/no-multi-comp */
+declare module '@vx/legend' {
+  import React from 'react';
+
+  export function LegendOrdinal(props: { [key: string]: any }): React.ReactElement;
```
williaster: just noting a mix of `ReactElement`/`ReactNode`, I'm not sure which is best TBH. May just have to experiment and see what happens 🔬

williaster (approved):
LGTM overall, had a couple thoughts but not sure we need to get too detailed here/nothing blocking.

@conglei are there any major differences you see vs how you've done this in Bighead?

packages/superset-ui-preset-chart-xy/types/@vx/responsive/index.d.ts
```diff
@@ -0,0 +1,6 @@
+declare module '@vx/responsive' {
+  import React from 'react';
+
+  // eslint-disable-next-line import/prefer-default-export
+  export class ParentSize extends React.Component<{}, {}> {}
```
conglei: We could make it more robust like:

```
  interface ParentSizeProps {
    children: (renderProps: { width: number; height: number }) => React.ReactNode;
  }

  const ParentSize: React.ComponentType<ParentSizeProps>;
```

conglei:
For me, I implemented the minimum declaration for `xy-chart` and `vx`. For `vx`, I borrowed the one from here `https://gist.github.com/techniq/e52d3f797c827b765ac67ff0cbe98ced` when it is available.  The most complete one for `data-ui` is `network` package.

conglei:
https://github.com/airbnb/bighead/blob/7441303aaa622641d071cd55668967c5c1865b6a/python/bighead/bighead_service/webapp/static/src/%40types/%40data-ui/xy-chart.d.ts 

This is the most related one with this PR, but it is also in a very early stage (with a lot of any lol)..

packages/superset-ui-preset-chart-xy/types/@data-ui/theme/index.d.ts
kristw: I was tracing the `@data-ui/theme` code and try to be as specific as I could. If the field is not defined in the `chartTheme` object, I did not add it to the type. Could easily add this to `SvgLabelTextStyle`

packages/superset-ui-preset-chart-xy/types/@vx/legend/index.d.ts
kristw: > React.Node. This represents any node that can be rendered in a React application. React.Node can be null, a boolean, a number, a string, a React element, or an array of any of those types recursively. If you need a return type for your component render() methods then you should use React.Node .

Hmm, maybe I should change the `ReactElement` to `ReactNode`

conglei (approved):

PR #13, opened 2019-03-12
 feat: add tooltip and layout components for charts

kristw:
🏆 Enhancements

* Add tooltip and layout components for charts
* Convert `WithLegend` to TypeScript

depends on #12


packages/superset-ui-preset-chart-xy/src/components/ChartFrame.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import { isDefined } from '@superset-ui/core';
+
+function checkNumber(input: any): input is number {
+  return isDefined(input) && typeof input === 'number';
+}
+
+interface Props {
```
williaster: `Props` should be a `type` not `interface`? 

looks like we have instances of `interface` in `superset-ui` but I thought we aligned on `type` because the styleguide recommends
> Use type for React components and constructs that can only be expressed with a type alias. Use interface for everything else. For a technical explanation of the differences and trade-offs, see [this post](https://medium.com/@martin_hotell/interface-vs-type-alias-in-typescript-2-7-2a8f1777af4c) about Interface vs Type alias in TypeScript 2.7.

packages/superset-ui-preset-chart-xy/src/components/ChartFrame.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+    const overflowX = checkNumber(contentWidth) && contentWidth > width;
+    const overflowY = checkNumber(contentHeight) && contentHeight > height;
+
+    if (overflowX || overflowY) {
+      return (
+        <div
+          style={{
+            height,
+            overflowX: overflowX ? 'scroll' : 'hidden',
```
williaster: sorry this got through in the prev component, `overflow` should never be `scroll` it should always be `auto`. `scroll` always shows a scrollbar whereas `auto` does not.

packages/superset-ui-preset-chart-xy/src/components/WithLegend.tsx
```diff
@@ -17,30 +17,23 @@
...
-  position: 'top',
-  hideLegend: false,
-};
+interface Props {
+  className: string;
+  width: number | string;
+  height: number | string;
+  legendJustifyContent: 'center' | 'flex-start' | 'flex-end';
+  position: 'top' | 'left' | 'bottom' | 'right';
+  renderChart: (dim: { width: number; height: number }) => ReactElement;
```
williaster: based on the other PR should this be `ReactNode`?

packages/superset-ui-preset-chart-xy/src/components/WithLegend.tsx
```diff
@@ -17,30 +17,23 @@
...
+  renderLegend: (params: { direction: string }) => ReactElement;
```
williaster: ditto `ReactNode`

packages/superset-ui-preset-chart-xy/src/components/tooltip/TooltipFrame.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+interface Props {
+  className?: string;
+  children: React.ReactChild;
```
williaster: I've had issues typing `children` correctly, if this works 👍 but curious if it works for `>1` child. (maybe supporting just 1 is fine, too)

packages/superset-ui-preset-chart-xy/src/components/tooltip/TooltipTable.tsx
```diff
@@ -0,0 +1,41 @@
+import React, { CSSProperties } from 'react';
+
+interface Props {
```
williaster: `type`

packages/superset-ui-preset-chart-xy/src/components/tooltip/TooltipTable.tsx
```diff
@@ -0,0 +1,41 @@
...
+  className?: string;
+  data: {
+    key: string;
+    keyStyle?: CSSProperties;
+    value: string | number;
+    valueStyle?: CSSProperties;
+  }[];
+}
+
+const defaultProps = {
```
williaster: any reason not to just move this into the `class`?

williaster (approved):
LGTM overall, I commented on a few TS things for `ReactElement` + `interface`.

packages/superset-ui-preset-chart-xy/src/components/ChartFrame.tsx
kristw: sgtm

packages/superset-ui-preset-chart-xy/src/components/tooltip/TooltipFrame.tsx
kristw: I will change this to `ReactNode`. It seems `ReactNode` also includes `ReactNodeArray` which sounds like support multiple

PR #15, opened 2019-03-13
feat: add encodeable utilities for chart

kristw:
🏆 Enhancements

### Define encoding config for a visualization

* Each visualization has one or more *channels* that can be encoded, such as `x`, `y`, `color`, `stroke`, fill. These channels configurations should be consistent/sharable across all visualizations. For example, `x` and `y` usually have scales and axes. Scales have `domain`, `range`. Axes have `label`, `format`. Color can support categorical, sequential or fixed color. All channels can support fixed value, field name in the data, or sometimes aggregates (haven't implemented the aggregate yet).
* Each encoding channel config is based on `vega-lite` encoding grammar. ([See line chart example](https://github.com/apache-superset/superset-ui-plugins/compare/kristw--line?expand=1#diff-89f0892b8e6df5ff19a42243e2254d6a), [vega-lite encoding documentation](https://vega.github.io/vega-lite/docs/encoding.html))
* Only adopt a small subset of `vega-lite` encoding grammar for now as we don’t have enough time (and may not intend) to provide 100% feature of `vega-lite`. 

### Implement TypeScript types for the encoding config.

* First attempt was using all types from `vega-lite` package (`XXXFieldDef`) directly but ended up getting too overwhelmed.
* Now adopting the primitive types from `vega-lite` then write higher-level types that has same name but only with the subset of the fields, so we can keep that of what we support and slowly adopt more in the future instead of taking it all now which our encoder does not know how to handle and also make it harder to implement the encoder. ([preview](https://github.com/apache-superset/superset-ui-plugins/compare/kristw--line?expand=1#diff-14e0823f4ab7530efbab63a9183f6d51))

### Write an encoder that parses the encoding config mention above and become a helper during visualization rendering.

* There is a base `Encoder` class and each visualization such as line chart will implement its own `Encoder` that extends from the base class. See: [base Encoder class](https://github.com/apache-superset/superset-ui-plugins/compare/kristw--line?expand=1#diff-16f12a4b98393d4c5370f6b564777c0b), [Line Chart's Encoder](https://github.com/apache-superset/superset-ui-plugins/compare/kristw--line?expand=1#diff-73daa5efbd07552f1824a8c865377070)

### Long shot

* Once all these encoding configs code are a bit more stable, it could be extracted to `@superset-ui/encodeable`.


packages/superset-ui-preset-chart-xy/src/encodeable/utils/isEnabled.ts
```diff
@@ -0,0 +1,11 @@
+export default function isEnabled(
+  config:
+    | {
+        [key: string]: any;
+      }
+    | boolean
+    | null
+    | undefined,
+) {
+  return config !== false || config !== null;
```
conglei: `config !== false && config !== null` ;

packages/superset-ui-preset-chart-xy/src/encodeable/types/data.ts
```diff
@@ -0,0 +1,8 @@
+export interface PlainObject {
+  [key: string]: any;
+}
+
+export interface Dataset {
+  keys: string[];
+  values: PlainObject[];
```
conglei: not sure if we can make sure that the key in the `PlainObject` will be always in the set of `key`? seems impossible.\

packages/superset-ui-preset-chart-xy/src/encodeable/BaseEncoder.ts
```diff
@@ -0,0 +1,98 @@
...
+      ...rest,
+      encoding: {
+        ...defaultEncoding,
+        ...encoding,
+      },
+    };
+  }
+
+  protected createChannel<ChannelName extends keyof Output>(
+    name: ChannelName,
```
conglei: just curious why do we need to pass `ChannelName` again as a parameter?

packages/superset-ui-preset-chart-xy/src/encodeable/types/fielddef.ts
```diff
@@ -0,0 +1,84 @@
+// eslint-disable no-unused-vars
```
williaster: should the name of these type files be `UpperCamelCase` like `FieldDef.ts` to be consistent with naming in e.g., `@superset-ui/chart` `types/`?

packages/superset-ui-preset-chart-xy/src/Line/Encoder.ts
```diff
@@ -0,0 +1,49 @@
...
+import { LooseSpec } from '../encodeable/types/spec';
+
+/**
+ * Define output type for each channel
+ */
+export interface Output {
+  x: number | null;
+  y: number | null;
+  color: string;
+  fill: boolean;
```
williaster, line 12: does a `boolean` mean it impossible to have different `color` / `fill`?

packages/superset-ui-preset-chart-xy/src/encodeable/BaseEncoder.ts
```diff
@@ -0,0 +1,98 @@
...
+  readonly legends: {
+    [key: string]: (keyof Encoding)[];
+  };
+
+  constructor(spec: LooseSpec<Encoding, Options>, defaultEncoding?: Encoding) {
+    this.spec = this.createFullSpec(spec, defaultEncoding);
+    this.channels = this.createChannels();
+    this.legends = {};
+
+    (Object.keys(this.channels) as (keyof Encoding)[])
```
williaster: Can you leave a comment for what this is doing? or could move to a helper function called `populateLegend` or something just to make it easier to read.

packages/superset-ui-preset-chart-xy/src/encodeable/BaseEncoder.ts
```diff
@@ -0,0 +1,98 @@
...
+          }
+        }
+      });
+  }
+
+  /**
+   * subclass can override this
+   */
+  protected createFullSpec(spec: LooseSpec<Encoding, Options>, defaultEncoding?: Encoding) {
+    if (defaultEncoding === undefined) {
```
williaster: should this be `typeof defaultEncoding === 'undefined'`?

packages/superset-ui-preset-chart-xy/src/encodeable/BaseEncoder.ts
```diff
@@ -0,0 +1,98 @@
...
+      encoding[name],
+      {
+        ...this.spec.options,
+        ...options,
+      },
+    );
+  }
+
+  /**
+   * subclass should override this
```
williaster: if this is expected, should you throw here instead?

packages/superset-ui-preset-chart-xy/src/encodeable/BaseEncoder.ts
williaster: yeah wondering if it's possible to clean up the signature `this.createChannel<name>(name, ...)`. I guess if consumer is not using TS we need the name parameter. so maybe we don't need the `ChannelName` generic variable and could instead infer from the name parameter on this line?

packages/superset-ui-preset-chart-xy/src/encodeable/ChannelEncoder.ts
```diff
@@ -0,0 +1,100 @@
...
+import parseScale from './parsers/parseScale';
+import parseGetter from './parsers/parseGetter';
+import parseFormat from './parsers/parseFormat';
+import parseAxis, { isXY } from './parsers/parseAxis';
+import isEnabled from './utils/isEnabled';
+import isDisabled from './utils/isDisabled';
+import { ChannelOptions } from './types/channel';
+import identity from './utils/identity';
+
+export default class ChannelEncoder<Def extends ChannelDef<V>, V extends Value = Value> {
```
williaster: I've really like the descriptive generic variables used so far, wondering if we could make `V` more readable? would help for readability below.

packages/superset-ui-preset-chart-xy/src/encodeable/ChannelEncoder.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  readonly axis?: PlainObject;
+  protected readonly getValue: (datum: PlainObject) => Value;
+  readonly scale?: ScaleOrdinal<string, V> | CategoricalColorScale | ((x: any) => V);
+  readonly formatter: Formatter;
+
+  readonly encodeValue: (value: any) => V;
+  readonly formatValue: (value: any) => string;
+
+  constructor(name: string, definition: Def, options: ChannelOptions = {} as ChannelOptions) {
```
williaster: is the duplicate `ChannelOptions ... as ChannelOptions` needed?

packages/superset-ui-preset-chart-xy/src/encodeable/ChannelEncoder.ts
```diff
@@ -0,0 +1,100 @@
...
+    this.name = name;
+    this.definition = definition;
+    this.options = options;
+
+    this.getValue = parseGetter(definition);
```
williaster: `parse` is a little confusing to me, would `extract` (`extractGetter`, `extractFormat`, etc) be more specific?

packages/superset-ui-preset-chart-xy/src/encodeable/ChannelEncoder.ts
```diff
@@ -0,0 +1,100 @@
...
+      this.formatValue = (value: any) => formatter(value);
+    } else if (formatter instanceof TimeFormatter) {
+      this.formatValue = (value: any) => formatter(value);
+    } else {
+      this.formatValue = formatter;
+    }
+
+    const scale = parseScale<V>(definition, options.namespace);
+    this.scale = scale;
+    if (scale === undefined) {
```
williaster: not sure if these should be `typeof undefined` comparisons like in JS, if TS complains this seems okay. 

I think the main reason for doing `typeof` is because theoretically you can set the value of `undefined` to anything you want, which made `typeof` more robust. but overrideing `undefined` seems unlikely in the ES6 world 😈

packages/superset-ui-preset-chart-xy/src/encodeable/parsers/parseAxis.ts
```diff
@@ -0,0 +1,35 @@
+import { cloneDeep } from 'lodash';
+import { Axis } from 'vega-lite/src/axis';
+import { ChannelDef, isPositionFieldDef, Formatter } from '../types/fielddef';
+import parseFormat from './parseFormat';
+import { PlainObject } from '../types/data';
+
+export function isXY(channelName: string) {
```
williaster: when used elsewhere I'd find this more readable if called `isXYChannel` because you don't see the argument var name is referring to channel.

packages/superset-ui-preset-chart-xy/src/encodeable/types/spec.ts
```diff
@@ -0,0 +1,13 @@
+export interface BaseOptions {
+  namespace?: string;
+}
+
+export interface LooseSpec<Encoding, Options extends BaseOptions = BaseOptions> {
```
williaster: `PartialSpec` is more clear to me/consistent with the idea of a `Partial`

packages/superset-ui-preset-chart-xy/src/encodeable/utils/isEnabled.ts
williaster: could also do `!isDisabled(config)`

packages/superset-ui-preset-chart-xy/src/encodeable/types/data.ts
williaster: I googled a bit and couldn't figure this out, worth asking in #typescript tho?

williaster (commented):
Looking great overall! some crazy TS in there 😎 

Thanks for the hard work on this!

packages/superset-ui-preset-chart-xy/src/Line/Encoder.ts
kristw: Umm, at this point yes. the fill is gradient from [stroke] `color` to white. 
I can make this `boolean | string`, which may lead to some ugly line charts with stroke and fill not getting along.

packages/superset-ui-preset-chart-xy/src/encodeable/utils/isEnabled.ts
kristw: Oh right.... stupid me

packages/superset-ui-preset-chart-xy/src/encodeable/BaseEncoder.ts
kristw: will change this to `abstract` function and make the class `abstract` as well

packages/superset-ui-preset-chart-xy/src/encodeable/ChannelEncoder.ts
kristw: can change to `typeof` for consistency and safety

packages/superset-ui-preset-chart-xy/src/encodeable/types/spec.ts
kristw: +1

packages/superset-ui-preset-chart-xy/src/encodeable/parsers/parseAxis.ts
kristw: +1

packages/superset-ui-preset-chart-xy/src/encodeable/types/fielddef.ts
kristw: sure

packages/superset-ui-preset-chart-xy/src/encodeable/types/data.ts
kristw: Can use typeguard to help. Not sure if possible to define solely as type without knowing the fields before hand.

```ts
export type PlainObject<Key extends string = string, Value extends any = any> = {
  [key in Key]: Value
};

export type Dataset<T extends string> = {
  keys: T[];
  values: Partial<PlainObject<T>>[];
}

function isDataset<T extends string>(input: Dataset<T>): Dataset<T> {
  return input;
}
```

image

kristw: image

kristw: mmm. this is more complicated. the guard only work for inline input, but if you pass a variable with type already it make `keys` `string[]` instead of `'a'|'b'|'c'[]` and the check does not work.

packages/superset-ui-preset-chart-xy/src/encodeable/BaseEncoder.ts
kristw: I tested and it can infer the correct `ChannelName` as type, so no need to define generic explicitly.

packages/superset-ui-preset-chart-xy/src/encodeable/BaseEncoder.ts
kristw: commented

kristw:
Addressed most of the comments.

packages/superset-ui-preset-chart-xy/package.json
```diff
@@ -29,21 +29,27 @@
...
     "@vx/scale": "^0.0.182",
     "@vx/shape": "^0.0.184",
+    "csstype": "^2.6.3",
     "d3-array": "^2.0.3",
     "d3-scale": "^2.2.2",
-    "prop-types": "^15.6.2"
+    "lodash": "^4.17.11",
+    "prop-types": "^15.6.2",
+    "reselect" : "^4.0.0",
+    "vega": "^5.2.0",
```
williaster, line 45: I didn't see any imports of `vega`, is this because it's a `vega-lite` dep? bundle size should be okay with the deep imports I'd think.

packages/superset-ui-preset-chart-xy/src/encodeable/types/FieldDef.ts
```diff
@@ -0,0 +1,90 @@
...
+import { NumberFormatter } from '@superset-ui/number-format';
+import { WithScale } from './Scale';
+import { WithXAxis, WithYAxis, WithAxis } from './Axis';
+import { WithLegend } from './Legend';
+
+export type Formatter = NumberFormatter | TimeFormatter | ((d: any) => string);
+
+// ValueDef is { value: xxx }
+
+// FieldDef is { field: 'fieldName', type: ... }
```
williaster: remove since it's defined below? `ValueDef` above seems useful if you are reading code but can't introspect the imported types :)

williaster (approved):
Overall this looks super clean and solid improvements vs the first round! 😍 🎉 I think none of my comments are blocking and that we could extend the `fill` option to accept a `string` in the future if need be.

thanks for all the hard work on this, pretty crazy TS here! ❤️ will approve from my end but still let @conglei comment/review if he wants!

conglei (approved):
Thanks for adding this! 🥂  It looks good to me!

packages/superset-ui-preset-chart-xy/package.json
kristw: Tried removing it and realize `vega-lite` needs it.

PR #17, opened 2019-03-13
fix: nvd3 line chart y axis bounds

kristw:
🐛 Bug Fix

* Previously the bound will not be applied if the extent of data is larger than the specified bound.
* Address Airbnb's PRODUCT-70657

🏠 Internal

* Add examples in storybook to demonstrate different settings for `yAxisBounds`
* Set default time format for storybook. 

image

@michellethomas @williaster @conglei @graceguo-supercat


williaster (approved):
lgtm! 

💟 the storybook! you could use the knobs to make those interactive, too 🤷‍♀️

kristw:
That's a great idea. Haven't used knobs before. Will try that for the new components.

PR #18, opened 2019-03-13
fix: dual line chart color consistency for secondary y axis

kristw:
🐛 Bug Fix

* Address Airbnb's PRODUCT-67932

> It appears that if a metric appears in the "metric_2" parameter of a dual-axis chart, it will not have the same color as a metric of the same name which is set in the "metric" parameter of a different chart, and visa versa. Practically, this means that the same metric will be colored inconsistently in a dashboard.

## Solution

This was due to `nvd3` appending `' (right axis)'` to category name before passing to color function. 

### Before 

image

### After

image

cc: @michellethomas @graceguo-supercat

### DIFF
--- a/packages/superset-ui-legacy-preset-chart-nvd3/src/utils.js
+++ b/packages/superset-ui-legacy-preset-chart-nvd3/src/utils.js
@@ -33,6 +33,7 @@ export function cleanColorInput(value) {
   // for superset series that should have the same color
   return String(value)
     .trim()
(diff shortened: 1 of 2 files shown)


michellethomas (approved):

michellethomas:
Looks like you may need to get someone on `apache-superset/embeddable-charts-team` to approve.

williaster (approved):

PR #19, opened 2019-03-14
fix: show only necessary tick labels on log scale

kristw:
🐛 Bug Fix

* For log scale, only show labels for 1, 10, 100, 1000, ... 
* Address Airbnb's PRODUCT-65125

### Before

image

### After

image

@michellethomas @graceguo-supercat


michellethomas (approved):
lgtm

williaster (approved):
bummer we can't write tests for all these great fixes! not worth it given the upcoming replacement vis's, tho :)

PR #23, opened 2019-03-19
fix: make pivot table handle null string

kristw:
🐛 Bug Fix

* make pivot table handle `'null'` string and do not try to format it as `NaN`

### Before

image

### After

image

related to: https://github.com/apache/incubator-superset/pull/4905
cc: @john-bodley


john-bodley:
Thanks @kristw, https://github.com/apache/incubator-superset/pull/4905 appreciates it.

williaster (approved):
solid 👍

PR #26, opened 2019-03-23
feat: line chart with revised encodeable utilities

kristw:
***tl;dr.* This PR add the first working version of the new Line Chart.**

## 🌕Code Overview

This PR adds two new `LineChart` plugins under the `preset-chart-xy` package. Both plugins point to the same `Chart` component but has different `ChartMetadata` and `transformProps`. The legacy-compatible plugin will have the same `ChartMetadata` with the `useLegacyApi` field set to `true`. The `transformProps` is the key difference as the legacy-compatible plugin converts legacy `formData` to the new `formData` while the new plugin's `transformProps` does very little.

image

### 📈 Line Chart

In `Line.tsx`, which is the main chart component. `Line` uses `XYChartLayout`, which is a cross-chart utility, to compute necessary margins and labelling strategy for the axes. It also uses `ChartLegend` to display legend. In addition, it has two chart-specific subcomponents. 

#### Encoder

The `Encoder` captures it visual encoding channels (`x`, `y`, `color`, `fill`, `strokeDasharray`). Each channel has a `ChannelType` (`X` | `Y` | `XBand` | `YBand` | `Color`, `Category` | `Text`) and `Output` (`string | boolean | number | null`). These two type parameters are used to generate `Encoding` type, which is a vega-lite-ish encoding grammar.

Once the `Encoder` is initialized and passed in the actual `encoding` config, it can provide utility functions for getting value from `datum`, encoding values, as well as things we usually want from `scale` or `axis`.

#### Tooltip 

This simply renders chart-specific tooltip.

### 📐 Chart Layout

A chart has multiple wrappers. At the top level `<WithLegend />` attaches legend to one side of the chart, splitting the container into two parts: legend and `ChartFrame`. Legend will take only necessary space and give the rest to `ChartFrame`. The actual chart is contained within the `ChartFrame` and usually has the same size. However, in some case when the chart has to guarantee minimum width and/or height to be usable and the `ChartFrame` is too small for that, the chart will be given the specified minimum width and/or height, making the `ChartFrame` scrollable instead.

IMG_4686

## 📘 Documentation

Update the storybook for line chart in `packages/superset-ui-plugins-demo/storybook/stories/preset-chart-xy`
* Basic line chart
* Legacy-compatible version
image
* Line chart with missing data
image
* Encoding dasharray and fill
image

## 🔮 Future Work

There are still work remaining after this PR. This PR is a major milestone but not completion.

- [ ]  Optimize to create `Encoder` only when necessary.
- [ ]  Make creating an `Encoder` class more convenient. 
- [x]  Merge `computeXAxisLayout()` and `computeYAxisLayout()` into `AxisAgent`.
- [ ]  Make `AxisAgent` derive config type (`XAxis | YAxis`) from channel type.
- [ ]  Handle data points surrounded by missing data.
- [ ]  Make storybook demo `tsx` work with knobs without complaining about types


packages/superset-ui-plugins-demo/storybook/stories/preset-chart-xy/Line/data/data.js
```diff
@@ -0,0 +1,226 @@
+/* eslint-disable sort-keys, no-magic-numbers */
```
williaster, line 1: I wonder if at some point it would be useful to abstract these out into a `@superset-ui/mock-data` package similar to `@vx/mock-data`. Can punt for now tho 🏉

packages/superset-ui-plugins-demo/storybook/stories/preset-chart-xy/Line/stories/basic.jsx
```diff
@@ -0,0 +1,56 @@
...
+            encoding: {
+              x: {
+                field: 'x',
+                type: 'temporal',
+                format: '%Y',
+                scale: {
+                  type: 'time',
+                },
+                axis: {
+                  orientation: 'bottom',
```
williaster: could abstract these into `knob`s for easy debugging and demo-ing what it does.

packages/superset-ui-plugins-demo/storybook/stories/preset-chart-xy/Line/index.js
```diff
@@ -1,8 +1,13 @@
+import { LineChartPlugin as LegacyLineChartPlugin } from '../../../../../superset-ui-preset-chart-xy/src/legacy';
 import { LineChartPlugin } from '../../../../../superset-ui-preset-chart-xy/src';
-import Stories from './Stories';
+import BasicStories from './stories/basic';
+import LegacyStories from './stories/legacy';
+import MissingStories from './stories/missing';
+import TimeShiftStories from './stories/timeShift';
 
-new LineChartPlugin().configure({ key: 'line2' }).register();
+new LegacyLineChartPlugin().configure({ key: 'v2-line/legacy' }).register();
```
williaster: you could make the `key`s constants / shareable between files so the stories and the registration are always in sync.

packages/superset-ui-plugins-demo/storybook/stories/preset-chart-xy/Line/stories/basic.jsx
```diff
@@ -0,0 +1,56 @@
+/* eslint-disable no-magic-numbers, sort-keys */
```
williaster, line 1: any reason not to make these `tsx`? (no strong opinion here just calling it out)

packages/superset-ui-preset-chart-xy/src/Line/Line.tsx
```diff
@@ -0,0 +1,236 @@
+/**
```
williaster: Is this ASF blurb needed? haven't seen in other files 🤔

packages/superset-ui-preset-chart-xy/src/Line/Line.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+type Props = {
+  className?: string;
+  width: number;
+  height: number;
+  margin?: Margin;
+  encoding: Encoding;
+  data: Dataset;
+  theme?: ChartTheme;
+} & typeof defaultProps;
```
williaster: is the `& typeof defaultProps;` needed?

packages/superset-ui-preset-chart-xy/src/Line/Line.tsx
```diff
@@ -0,0 +1,236 @@
...
+}
+
+export interface SeriesValue {
+  x: Outputs['x'];
+  y: Outputs['y'];
+  data: PlainObject;
+  parent: Series;
+}
+
+class LineChart extends React.PureComponent<Props, {}> {
```
williaster: no need to pass the `state = {}` type if you aren't using it (that's the default already I believe)

packages/superset-ui-preset-chart-xy/src/Line/Line.tsx
```diff
@@ -0,0 +1,236 @@
...
+      <WithLegend
+        className={`superset-chart-line ${className}`}
+        width={width}
+        height={height}
+        position="top"
+        renderLegend={() => (
+          // eslint-disable-next-line react/jsx-props-no-multi-spaces
+          <ChartLegend<ChannelTypes, Outputs, Encoding> data={data} encoder={this.encoder} />
+        )}
+        renderChart={parent => this.renderChart(parent)}
```
williaster: no need for an inline function since you are passing the arg as-is `renderChart={this.renderChart}`? (you do need to bind `renderChart` tho in the `constructor`)

packages/superset-ui-preset-chart-xy/src/Line/Line.tsx
```diff
@@ -0,0 +1,236 @@
...
+        hideLegend={!this.encoder.hasLegend()}
```
williaster: could `renderLegend` just conditionally return `null` instead of adding another prop? not a huge opinion on this, could see an argument either way.

packages/superset-ui-preset-chart-xy/src/Line/Line.tsx
```diff
@@ -0,0 +1,236 @@
...
+        x: this.encoder.channels.x.encode(v),
+        y: this.encoder.channels.y.encode(v),
+        data: v,
+        parent: series,
+      }));
+
+      return series;
+    });
+
+    const children = flatMap(
```
williaster: Couple thoughts
- this is not super readable, you could possibly separate out into `const FilledSeries = ...` and `const LineSeries = ...`and then call `flatMap` instead of chaining everything together.

- Does this duplicate components for filled series? you filter for `filled` in the first `map`, but _don't_ filter for `unfilled` in the second. The area series do have a stroke, so should be able to just have one series for them.

packages/superset-ui-preset-chart-xy/src/Line/Line.tsx
```diff
@@ -0,0 +1,236 @@
...
+              strokeWidth={1.5}
+            />,
+          ];
+        }),
+    ).concat(
+      allSeries.map(series => (
+        <LineSeries
+          key={series.key}
+          seriesKey={series.key}
+          animated
```
williaster: this isn't a valid prop in the `xy-chart` package I think, just `histogram` IIRC.

packages/superset-ui-preset-chart-xy/src/Line/Line.tsx
```diff
@@ -0,0 +1,236 @@
...
+              strokeDasharray=""
+              showHorizontalLine={false}
+              circleFill={(d: SeriesValue) =>
+                d.y === tooltipData.datum.y ? d.parent.color : '#fff'
+              }
+              circleSize={(d: SeriesValue) => (d.y === tooltipData.datum.y ? 6 : 4)}
+              circleStroke={(d: SeriesValue) =>
+                d.y === tooltipData.datum.y ? '#fff' : d.parent.color
+              }
+              circleStyles={{ strokeWidth: 1.5 }}
```
williaster: could make a `const`

packages/superset-ui-preset-chart-xy/src/Line/legacy/index.ts
```diff
@@ -0,0 +1,31 @@
+/**
```
williaster: ditto ASF blurb ❓

packages/superset-ui-preset-chart-xy/src/Line/legacy/transformProps.ts
```diff
@@ -0,0 +1,67 @@
...
+  values: {
+    x: number;
+    y: number;
+  }[];
+}
+
+export default function transformProps(chartProps: ChartProps) {
+  const { width, height, formData, payload } = chartProps;
+  const { colorScheme, xAxisLabel, xAxisFormat, yAxisLabel, yAxisFormat } = formData;
+  const data: DataRow[] = payload.data as DataRow[];
```
williaster: do you need both `: DataRow[]` and `as DataRow[]`?

packages/superset-ui-preset-chart-xy/src/components/ChartLegend.tsx
```diff
@@ -0,0 +1,91 @@
...
+  Props<AbstractEncoder<ChannelTypes, Outputs, Encoding, Options>>,
+  {}
+> {
+  render() {
+    const { data, encoder } = this.props;
+
+    const legends = Object.keys(encoder.legends).map((field: string) => {
+      const channelNames = encoder.legends[field];
+      const channelEncoder = encoder.channels[channelNames[0]];
+      const keySet = new Set();
```
williaster: you can initialize a `Set` from an array to skip a line `new Set(data.values.map(channelEncoder.get))`

packages/superset-ui-preset-chart-xy/src/components/ChartLegend.tsx
```diff
@@ -0,0 +1,91 @@
...
+        range: domain.map((key: string) => channelEncoder.encodeValue(key)),
+      });
+
+      return (
+        <div key={field} style={LEGEND_CONTAINER_STYLE}>
+          <LegendOrdinal scale={scale} labelFormat={channelEncoder.formatValue}>
+            {(labels: Label[]) => (
+              <div style={{ display: 'flex', flexDirection: 'row', flexWrap: 'wrap' }}>
+                {labels.map((label: Label) => {
+                  const size = 8;
```
williaster: could be outside the loop 🤷‍♀️

packages/superset-ui-preset-chart-xy/src/components/ChartLegend.tsx
```diff
@@ -0,0 +1,91 @@
...
+                  );
+                })}
+              </div>
+            )}
+          </LegendOrdinal>
+        </div>
+      );
+    });
+
+    return <React.Fragment>{legends}</React.Fragment>;
```
williaster, line 89: guess we only added the `<>` lint rule internally, can make these consistent later.

packages/superset-ui-preset-chart-xy/src/encodeable/AbstractEncoder.ts
```diff
@@ -1,52 +1,80 @@
...
-    (Object.keys(this.channels) as (keyof Encoding)[])
-      .map((key: keyof Encoding) => this.channels[key])
+    this.legends = {};
+    channelNames
+      .map((name: ChannelName) => this.channels[name])
       .filter(c => c.hasLegend())
       .forEach(c => {
         if (isFieldDef(c.definition)) {
-          const key = c.name as keyof Encoding;
+          const name = c.name as ChannelName;
```
williaster, line 72: nit -- object destructuring (I think this is a lint rule but only for react props)

packages/superset-ui-preset-chart-xy/src/encodeable/parsers/extractScale.ts
```diff
@@ -1,38 +1,241 @@
...
+
+  if (typeof scale !== 'undefined') {
+    if (scale.domain && typeof config.domain !== 'undefined') {
+      scale.domain(config.domain as any[]);
+    }
+    if (scale.range && typeof config.range !== 'undefined') {
+      scale.range(config.range as any[]);
+    }
+    if ('nice' in scale && scale.nice) {
+      if (!(typeof config.nice === 'boolean' && config.nice === false)) {
```
williaster: could this just be `config.nice !== false`?

packages/superset-ui-preset-chart-xy/src/utils/XYChartLayout.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+interface Input {
+  width: number;
+  height: number;
+  minContentWidth?: number;
+  minContentHeight?: number;
+  margin: Margin;
+  xEncoder: ChannelEncoder<XFieldDef>;
+  yEncoder: ChannelEncoder<YFieldDef>;
+  children: React.ReactElement[];
```
williaster: `React.ReactNode`?

packages/superset-ui-preset-chart-xy/src/utils/XYChartLayout.tsx
```diff
@@ -0,0 +1,185 @@
...
+    }
+    this.chartWidth = isOverFlowX ? chartWidth + OVERFLOW_MARGIN : chartWidth;
+    this.chartHeight = isOverFlowY ? chartHeight + OVERFLOW_MARGIN : chartHeight;
+    this.containerWidth = width;
+    this.containerHeight = height;
+    this.margin = finalMargin;
+  }
+
+  recommendLabelAngle(xOrientation: 'top' | 'bottom' = 'bottom') {
+    if (!this.yLayout) {
```
williaster: could this condition just be included as part of the condition below for `+40`?

packages/superset-ui-preset-chart-xy/src/utils/XYChartLayout.tsx
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    return (this.yLayout.orientation === 'right' && xOrientation === 'bottom') ||
+      (this.yLayout.orientation === 'left' && xOrientation === 'top')
+      ? 40
+      : -40;
+  }
+
+  createChartWithFrame(
+    renderChart: (input: { width: number; height: number }) => React.ReactElement,
```
williaster: `React.ReactNode`?

packages/superset-ui-preset-chart-xy/src/utils/XYChartLayout.tsx
```diff
@@ -0,0 +1,185 @@
...
+        width={this.containerWidth}
+        height={this.containerHeight}
+        contentWidth={this.chartWidth}
+        contentHeight={this.chartHeight}
+        renderContent={renderChart}
+      />
+    );
+  }
+
+  createXAxis(props?: PlainObject) {
```
williaster: would it be appropriate to call these `renderXXX` since they return `React` nodes?

packages/superset-ui-preset-chart-xy/src/utils/XYChartLayout.tsx
williaster: I might be missing something obvious but why is `0` not the default rotation?

packages/superset-ui-preset-chart-xy/src/utils/XYChartLayout.tsx
```diff
@@ -0,0 +1,185 @@
...
+      finalMargin.right += OVERFLOW_MARGIN;
+    }
+    this.chartWidth = isOverFlowX ? chartWidth + OVERFLOW_MARGIN : chartWidth;
+    this.chartHeight = isOverFlowY ? chartHeight + OVERFLOW_MARGIN : chartHeight;
+    this.containerWidth = width;
+    this.containerHeight = height;
+    this.margin = finalMargin;
+  }
+
+  recommendLabelAngle(xOrientation: 'top' | 'bottom' = 'bottom') {
```
williaster: naming this `recommendXLabelAngle` or `getDefaultXLabelAngle` could clarify that it is only used for `XAxis`.

packages/superset-ui-preset-chart-xy/src/utils/computeXAxisLayout.ts
```diff
@@ -1,21 +1,34 @@
...
+import { CSSProperties } from 'react';
 
 export default function computeXAxisLayout({
   axisLabelHeight = 20,
   axisWidth,
   gapBetweenAxisLabelAndBorder = 8,
   gapBetweenTickAndTickLabel = 4,
   gapBetweenTickLabelsAndAxisLabel = 4,
-  labellingStrategy = 'auto',
+  labelAngle = -40,
```
williaster: ditto question about defaulting to `0`

packages/superset-ui-preset-chart-xy/src/utils/createTickComponent.tsx
```diff
@@ -1,13 +1,17 @@
 /* eslint-disable no-magic-numbers */
 
-import React from 'react';
-import PropTypes from 'prop-types';
+import React, { CSSProperties } from 'react';
 
 export default function createTickComponent({
   labellingStrategy,
   orientation = 'bottom',
   rotation = 40,
```
williaster: should this be a constant? there are at least 3 files that set a default, and some don't match so it's tricky to figure out what the actual default will be.

packages/superset-ui-preset-chart-xy/src/utils/createTickComponent.tsx
```diff
@@ -16,28 +20,26 @@ export default function createTickComponent({
...
-      x: PropTypes.number.isRequired,
-      y: PropTypes.number.isRequired,
-    };
-    const defaultProps = {
-      dy: null,
-      formattedValue: '',
-    };
-
-    const TickComponent = ({ x, y, dy, formattedValue, ...textStyle }) => (
+    const TickComponent = ({
```
williaster: I realize this component wasn't added here, but just to understand why we didn't leverage the `@vx/text` component for auto wrapping / easy rotation etc., is it mostly that the manual calculations are needed to move the margins dynamically?

packages/superset-ui-preset-chart-xy/types/@vx/legend/index.d.ts
```diff
@@ -2,15 +2,15 @@
 declare module '@vx/legend' {
   import React from 'react';
 
-  export function LegendOrdinal(props: { [key: string]: any }): React.ReactNode;
+  export function LegendOrdinal(props: { [key: string]: any }): React.ReactElement;
```
williaster: didn't like `ReactNode`? (you can ignore/resolve my other suggestions of `ReactNode` if they don't work, was mostly trying to identify spots where we deviate from `Node`)

williaster (commented):
Looking great overall, thanks for all the effort here! 💛 💜

packages/superset-ui-plugins-demo/storybook/stories/preset-chart-xy/Line/data/data.js
kristw: I create an issue for this. https://github.com/apache-superset/superset-ui-plugins/issues/28

packages/superset-ui-preset-chart-xy/src/utils/XYChartLayout.tsx
kristw: 👍

packages/superset-ui-preset-chart-xy/src/Line/Line.tsx
kristw: 👍

packages/superset-ui-preset-chart-xy/src/Line/Line.tsx
kristw: 🤦‍♂️ my bad copy-pasting.

packages/superset-ui-preset-chart-xy/src/Line/legacy/transformProps.ts
kristw: good point. removed the first one.

packages/superset-ui-preset-chart-xy/src/components/ChartLegend.tsx
kristw: We can do one pr to enable that and run `eslint-fix` the entire repo.

kristw: for tracking https://github.com/apache-superset/superset-ui-plugins/issues/29

packages/superset-ui-preset-chart-xy/src/encodeable/AbstractEncoder.ts
kristw: Also need to cast its type. Can I do that + destructure?

packages/superset-ui-preset-chart-xy/src/Line/Line.tsx
kristw: This will ensure `margin` and `theme`'s types are not `undefined`

packages/superset-ui-preset-chart-xy/src/Line/Line.tsx
kristw: 🤦‍♂️ true...

packages/superset-ui-preset-chart-xy/src/Line/Line.tsx
kristw: 👍 on the duplicate `Area` and `Line`. 
I have a habit of chaining when applicable in general

packages/superset-ui-preset-chart-xy/types/@vx/legend/index.d.ts
kristw: image

kristw: These require `ReactElement` because expectation is a constructor. `ReactNode` is more suitable for child which can be `string | number | null | undefined`

packages/superset-ui-preset-chart-xy/src/utils/XYChartLayout.tsx
kristw: There are two fields `labelOverlap` and `labelAngle`. 

`labelOverlap` can be `'auto' | 'flat' | 'rotate'`

`labelAngle` is used only when `labelOverlap` 
* is `'rotate'` 
* is `'auto'` and then the heuristics determines that there are overlaps.

packages/superset-ui-plugins-demo/storybook/stories/preset-chart-xy/Line/stories/basic.jsx
kristw: I converted it to `tsx` then add knobs. Then it starts complaining for missing types definition for external module. I renamed it back to `jsx` and will take care of the `tsx` and external type def issue in a follow-up pr.

packages/superset-ui-preset-chart-xy/src/Line/Line.tsx
conglei: I like this pattern!

packages/superset-ui-preset-chart-xy/src/Line/Line.tsx
```diff
@@ -0,0 +1,226 @@
...
+}
+
+export interface SeriesValue {
+  x: Outputs['x'];
+  y: Outputs['y'];
+  data: PlainObject;
+  parent: Series;
+}
+
+const CIRCLE_STYLE = { strokeWidth: 1.5 };
```
conglei, line 55: does it mean users cannot customize the style? If the chart is tailored specially for superset, I won't see it will be an issue. But just was thinking of the customization part.

packages/superset-ui-preset-chart-xy/src/components/ChartLegend.tsx
```diff
@@ -0,0 +1,91 @@
...
+      const domain = Array.from(new Set(data.values.map(channelEncoder.get)));
+      const scale = scaleOrdinal({
+        domain,
+        range: domain.map((key: string) => channelEncoder.encodeValue(key)),
+      });
+
+      return (
+        <div key={field} style={LEGEND_CONTAINER_STYLE}>
+          <LegendOrdinal scale={scale} labelFormat={channelEncoder.formatValue}>
+            {(labels: Label[]) => (
```
conglei, line 59: Have we done the test with extremely large number of legends here? So far, there are a lot of chart in Superset that has more than I'd say 20 legends. At least we need to make sure the chart can be rendered.

packages/superset-ui-preset-chart-xy/src/components/WithLegend.tsx
```diff
@@ -29,34 +11,32 @@ type Props = {
   legendJustifyContent: 'center' | 'flex-start' | 'flex-end';
   position: 'top' | 'left' | 'bottom' | 'right';
   renderChart: (dim: { width: number; height: number }) => ReactNode;
-  renderLegend: (params: { direction: string }) => ReactNode;
-  hideLegend: boolean;
+  renderLegend?: (params: { direction: string }) => ReactNode;
```
conglei, line 22: I saw we have two different ways to define Props with defaultProps. Can we make the styles consistent? I do like the way you defined before, with `& typeof defaultProps`.

packages/superset-ui-preset-chart-xy/src/encodeable/parsers/extractScale.ts
```diff
@@ -1,38 +1,246 @@
...
+    | ScalePoint<{ toString(): string }>
+    | ScaleBand<{ toString(): string }>;
+}
+
+export interface ScaleTypeToD3ScaleType<Output> {
+  [ScaleType.LINEAR]: ScaleLinear<Output, Output>;
+  [ScaleType.LOG]: ScaleLogarithmic<Output, Output>;
+  [ScaleType.POW]: ScalePower<Output, Output>;
+  [ScaleType.SQRT]: ScalePower<Output, Output>;
+  [ScaleType.SYMLOG]: ScaleLogarithmic<Output, Output>;
```
conglei, line 58: wow, you really considered all the possibilities ! 👍

conglei (approved):
Awesome work!! Just left some comments on 

1) Customizability issue'

2) the style Props and defaultProps.  

Both are nit :)

packages/superset-ui-preset-chart-xy/src/encodeable/parsers/extractScale.ts
kristw: image

packages/superset-ui-preset-chart-xy/src/Line/Line.tsx
kristw: This is only for the highlighted circle when using crosshair. May have to think more carefully about how to expose customization.

packages/superset-ui-preset-chart-xy/src/components/ChartLegend.tsx
kristw: `LEGEND_CONTAINER_STYLE` caps the legend with `maxHeight`.

packages/superset-ui-preset-chart-xy/src/components/WithLegend.tsx
kristw: Yes, fixed.

PR #27, opened 2019-03-24
style: add % to total to area chart tooltip

thomaswang-shift:
This PR formats the tooltip for the area chart to include a new column for displaying `% to total`.

### DIFF
--- a/packages/superset-ui-legacy-preset-chart-nvd3/src/utils.js
+++ b/packages/superset-ui-legacy-preset-chart-nvd3/src/utils.js
@@ -133,7 +133,36 @@ export function generateRichLineTooltipContent(d, timeFormatter, valueFormatter)
   });
   tooltip += '</tbody></table>';
 
-  return tooltip;
+  return dompurify.sanitize(tooltip);
+}
+
+export function generateAreaChartTooltipContent(d, timeFormatter, valueFormatter) {
+  const total = d.series[d.series.length - 1].value;
+  let tooltip = '';
+  tooltip +=
+    "<table><thead><tr><td colspan='4'>" +
+    `<strong class='x-value'>${timeFormatter(d.value)}</strong>` +
+    '</td></tr></thead><tbody>' +
+    '<tr><td></td><td>Category</td><td>Value</td><td>% to total</td></tr>';
+  d.series.forEach(series => {
+    const key = getFormattedKey(series.key, true);
+    let trClass = '';
+    if (series.highlight) {
+      trClass = 'superset-legacy-chart-nvd3-tr-highlight';
+    } else if (series.key === 'TOTAL') {
+      trClass = 'superset-legacy-chart-nvd3-tr-total';
+    }
+    tooltip +=
+      `<tr class="${trClass}" style="border-color: ${series.color}">` +
+      `<td style="color: ${series.color}">${series.key === 'TOTAL' ? '' : '&#9724;'}</td>` +
+      `<td>${key}</td>` +
+      `<td>${valueFormatter(series.value)}</td>` +
+      `<td>${((100 * series.value) / total).toFixed(2)}%</td>` +
+      '</tr>';
+  });
+  tooltip += '</tbody></table>';
+
+  return dompurify.sanitize(tooltip);
 }
 
 export function generateMultiLineTooltipContent(d, xFormatter, yFormatters) {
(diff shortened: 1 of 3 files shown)


packages/superset-ui-legacy-preset-chart-nvd3/src/utils.js
```diff
@@ -132,7 +132,29 @@ export function generateRichLineTooltipContent(d, timeFormatter, valueFormatter)
...
   tooltip += '</tbody></table>';
+  return tooltip;
+}
 
+export function generateAreaChartTooltipContent(d, timeFormatter, valueFormatter) {
+  const total = d.series[d.series.length - 1].value;
+  let tooltip = '';
+  tooltip +=
+    "<table><thead><tr><td colspan='3'>" +
+    `<strong class='x-value'>${timeFormatter(d.value)}</strong>` +
```
mistercrunch, line 144: Should we sanitize / "html escape" what gets evaluated throughout this snipet?

dorq:
From @datability-io ...
image (6)

dorq:
This is looking great!  A few questions:
- Does this PR remove the colored blocks next to the values?  From another area chart in production now:
image
- Does this PR remove the bold and line styling on the hover value and the total?
- Do we need to include "category" and "value" as column headers?  If we do, can we capitalize them?  I'm assuming these names are standard across all charts.
- Can we left-align the date at the top with the first column below it?

packages/superset-ui-legacy-preset-chart-nvd3/src/utils.js
xtinec: Is this what you mean @mistercrunch ?
https://stackoverflow.com/questions/7381974/which-characters-need-to-be-escaped-in-html
I wonder if we have a utility function for this.

mistercrunch: We use `dompurify.sanitize` in other places. Basically preventing XSS. In theory everything that is inputed by a user (webform) or comes from an external database should be sanitize when put into raw html... If `d.value` contains `<script>doBadStuff();</script>` it shouldn't get evaluated, it should just show up in the tooltip.
https://github.com/cure53/DOMPurify#what-does-it-do

mistercrunch: A bit more about this. React "sanitizes" ("html escapes") all strings you put into it, everywhere by default. If you want user input to get rendered as HTML for instance, you'll have to use `dangerouslySetInnerHTML` https://zhenyong.github.io/react/tips/dangerously-set-inner-html.html

I think Jinja too, unless you mark something as safe.

So anytime we see html strings getting concatenated with user-provided content, we need to watch out for XSS.

thomaswang-shift: Well explained! @mistercrunch

kristw:
Please make sure your code pass travis CI

thomaswang-shift:
This is the result of the latest change.  We should be good to go on this PR.

image

thomaswang-shift:
@kristw, it looks like the Travis CI failure is not related to the change in this PR and it's also failing on master.

dorq:
LGTM.  Thanks, @datability-io !

kristw:
Your changes fails linting.

image

packages/superset-ui-legacy-preset-chart-nvd3/src/utils.js
```diff
@@ -133,7 +133,36 @@ export function generateRichLineTooltipContent(d, timeFormatter, valueFormatter)
...
+  const total = d.series[d.series.length - 1].value;
+  let tooltip = '';
+  tooltip +=
+    "<table><thead><tr><td colspan='4'>" +
+    `<strong class='x-value'>${timeFormatter(d.value)}</strong>` +
+    '</td></tr></thead><tbody>' +
+    '<tr><td></td><td>Category</td><td>Value</td><td>% to total</td></tr>';
+  d.series.forEach(series => {
+    const key = getFormattedKey(series.key, true);
+    let trCls = '';
```
kristw: Please use more specific variable name. Perhaps `trClass`

kristw:
something is wrong. Please make sure you rebase from upstream master correctly.

thomaswang-shift:
🔨

kristw:
@datability-io This PR modifies 64 files. I don't think that is your intention. 
image

xtinec (approved):
Looks good!

kristw (approved):

PR #31, opened 2019-03-29
feat: update line chart tooltip and use selector

kristw:
🏆 Enhancements

* Use `selector` to avoid recreating `Encoder` 
* Update Line Chart tooltip to use line mark to indicate encoding and change text back to black. 

image

image

### DIFF
--- a/packages/superset-ui-plugins-demo/tsconfig.json
+++ b/packages/superset-ui-plugins-demo/tsconfig.json
@@ -1,23 +1,28 @@
 {
   "compilerOptions": {
+    "allowJs": true,
+    "allowSyntheticDefaultImports": true,
     "baseUrl": ".",
     "outDir": "./dist",
-    "module": "commonjs",
-    "target": "es5",
-    "lib": ["es6", "dom"],
-    "sourceMap": true,
-    "allowJs": true,
-    "jsx": "react",
-    "moduleResolution": "node",
+    "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
+    "importHelpers": true,
+    "jsx": "react",
+    "lib": ["dom", "esnext"],
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "noEmitOnError": true,
     "noImplicitReturns": true,
     "noImplicitThis": true,
     "noImplicitAny": true,
-    "importHelpers": true,
+    "noUnusedLocals": true,
+    "pretty": true,
+    "removeComments": false,
     "strictNullChecks": true,
     "suppressImplicitAnyIndexErrors": true,
-    "noUnusedLocals": true,
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "sourceMap": true,
+    "target": "esnext"
   },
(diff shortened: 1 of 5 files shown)


packages/superset-ui-plugins-demo/tsconfig.json
```diff
@@ -1,23 +1,28 @@
 {
```
williaster, line 1: is this file not generated by beemo?

packages/superset-ui-preset-chart-xy/src/Line/createTooltip.tsx
```diff
@@ -31,13 +33,24 @@ export default function createTooltip(encoder: Encoder, allSeries: Series[]) {
...
+                .map(({ key, color, strokeDasharray }) => ({
                   key,
-                  keyStyle: {
-                    color,
-                    fontWeight: series[key] === datum ? 600 : 200,
-                  },
-                  value: encoder.channels.y.formatValue(series[key].y),
+                  keyColumn: (
+                    <>
+                      <svg width="12" height="8" style={MARK_STYLE}>
```
williaster, line 40: should this glyph not match the legend? that seems ideal. it's tricky with the dashed line, tho. In that case I wonder if we should update the legend to be lines instead of circles or squares. wdyt?

packages/superset-ui-preset-chart-xy/src/components/tooltip/TooltipTable.tsx
```diff
@@ -1,11 +1,12 @@
...
 
 type Props = {
   className?: string;
   data: {
-    key: string;
+    key: string | number;
+    keyColumn: ReactNode;
     keyStyle?: CSSProperties;
-    value: string | number;
+    valueColumn: ReactNode;
```
williaster, line 9: are there cases where this is **not** `string/number/null`? `ReactNode` seems strange to have inside an array of data objects.

williaster (approved):
Had a couple comments, looks good overall!

packages/superset-ui-plugins-demo/tsconfig.json
kristw: Hmm... true. The one in `demo` is somehow not generated and is also checked in.

packages/superset-ui-preset-chart-xy/src/components/tooltip/TooltipTable.tsx
kristw: I'm trying to be flexible in case extra formatting is needed or there are unforeseen thing we want to put in the value column. Also to make it same with `keyColumn`

packages/superset-ui-preset-chart-xy/src/Line/createTooltip.tsx
kristw: Perhaps should update the legend to accept mark type and display line as well.

kristw: Making the legend customizable seems like a good chunk of work itself. I will make it a follow-up PR.

PR #32, opened 2019-03-29
chore: add eslint react fragment rule

### LINKED ISSUE #29: Use <> instead of <React.Fragment>
**Is your feature request related to a problem? Please describe.**
Consistent usage of `<>` vs `<React.Fragment>`

**Describe the solution you'd like**
Enable `eslint` rule to use `<>` and apply `eslint --fix` to entire repo 

**Reported by**
@williaster

kristw:
🏠 Internal

* Enforce using `<>` instead of `<React.Fragment>`
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-fragments.md

* `Fragment` is only available after `react` `16.2` so update the requirement for packages that use `Fragment` as well. 

* Close #29

### DIFF
--- a/packages/superset-ui-preset-chart-xy/package.json
+++ b/packages/superset-ui-preset-chart-xy/package.json
@@ -53,6 +53,6 @@
     "@superset-ui/number-format": "^0.10.0",
     "@superset-ui/time-format": "^0.10.0",
     "@superset-ui/translation": "^0.10.0",
-    "react": "^15 || ^16"
+    "react": "^15 || ^16.2"
   }
 }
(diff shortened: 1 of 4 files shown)


packages/superset-ui-preset-chart-xy/package.json
```diff
@@ -53,6 +53,6 @@
     "@superset-ui/number-format": "^0.10.0",
     "@superset-ui/time-format": "^0.10.0",
     "@superset-ui/translation": "^0.10.0",
-    "react": "^15 || ^16"
+    "react": "^15 || ^16.2"
```
williaster: I'm confused about this, should it not be `^16.2`?

williaster (approved):
lgtm after the React dep change!

PR #35, opened 2019-04-01
build: use shared commit config

kristw:
🏠 Internal

command line tool for creating commit as well as dependencies for `commitlint` are separated into package `@superset-ui/commit-config`

This PR update this repo to use `@superset-ui/commit-config`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@superset-ui/plugins-monorepo",
-  "version": "0.0.0-next",
+  "version": "0.0.0-master",
   "description": "Superset UI Plugins",
   "private": true,
   "scripts": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,14 +1,15 @@
 {
   "name": "@superset-ui/plugins-monorepo",
-  "version": "0.0.0",
+  "version": "0.0.0-next",
```
williaster: what does this do?

williaster (approved):
Looks like mostly changelog changes, but moving the config to a separate repo is a good call! 👍 👍

package.json
kristw: umm, not really doing anything. 

When running the changelog command and there are new commits after the previous release. The `conventional-changelog` cli takes the version from the root `package.json` to use as section header for the new commits. I was confused a few times why there is a section for `0.0.0` and was going to change `0.0.0` to `master`, but it is not `semver` valid, so I change to `0.0.0-master` instead.

PR #36, opened 2019-04-01
fix: exception thrown for charts without a x-axis

khtruong:
🐛 Bug Fix

We need to check if a chart has a x-axis before defining the tick format.

### DIFF
--- a/packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
+++ b/packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
@@ -495,15 +495,17 @@ function nvd3Vis(element, props) {
     if (chart.x2Axis && chart.x2Axis.tickFormat) {
       chart.x2Axis.tickFormat(xAxisFormatter);
     }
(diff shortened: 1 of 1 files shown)


packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
```diff
@@ -498,7 +498,7 @@ function nvd3Vis(element, props) {
     const isXAxisString = isVizTypes(['dist_bar', 'box_plot']);
     if (!isXAxisString && chart.xAxis && chart.xAxis.tickFormat) {
       chart.xAxis.tickFormat(xAxisFormatter);
-    } else {
+    } else if (chart.xAxis && chart.xAxis.tickFormat) {
```
kristw: Could we combine the check with the condition above?

```js
if (chart.xAxis && chart.xAxis.tickFormat) {
  if(isXAxisString) {
  } else {
  }
}
```

khtruong: Done!

kristw (approved):

PR #40, opened 2019-04-04
fix(legacy-preset-chart-nvd3): stacked bar charts labels

betodealmeida:
🐛 Bug Fix

Labels are not being displayed on stacked bar charts with some of the groups disabled in the legend, because the number of series being displayed is being computed incorrectly. I fixed it by ignoring disabled series, and added a new story to test it.


khtruong:
👍

kristw (approved):

PR #42, opened 2019-04-07
fix: remove sticky tooltip when query returns no data in explore view

xtinec:
🐛 Bug Fix

Tap into the newly added `willUnmount` hook in https://github.com/apache-superset/superset-ui/pull/131 to clean up sticky tooltip that nvd3 adds to `body` directly. This removes such tooltips that were left behind in the explore view with the new query returns 'no data'.
image

### DIFF
--- a/packages/superset-ui-legacy-preset-chart-nvd3/src/ReactNVD3.js
+++ b/packages/superset-ui-legacy-preset-chart-nvd3/src/ReactNVD3.js
@@ -18,5 +18,10 @@
  */
 import { reactify } from '@superset-ui/chart';
 import Component from './NVD3Vis';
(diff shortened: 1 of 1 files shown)


williaster (approved):
👍 Is there any difference between hide vs remove? Or hide _does_ remove?

xtinec:
It does either. The parameter `shouldRemove` determines whether to hide or to remove. 🙂 https://github.com/apache-superset/superset-ui-plugins/blob/789596f8c5b9a91da2e9bab7d397a55dc934b8fa/packages/superset-ui-legacy-preset-chart-nvd3/src/utils.js#L254

PR #43, opened 2019-04-08
chore: rename willUnmount hook to match the name in `superset-ui/chart`

xtinec:
🏠 Internal
rename willUnmount hook to match the name in `superset-ui/chart`


williaster (approved):
oi sorry about that, thanks for making the change 🙏

PR #56, opened 2019-04-12
fix: extract css per package to style.css

kristw:
An alternative solution to #55


khtruong (approved):

kristw:
mmm. still broken when building production storybook

PR #57, opened 2019-04-12
fix: list css as side effects

kristw:
🐛 Bug Fix

**Problem**
related to apache/incubator-superset#7058

The root cause ends up being the `sideEffects: true` that is set in each `package.json`. `webpack` in `production` mode uses this to determine if the package has any side effects and when it is declared `false`, it (via treeshaking) drops the file that does not export anything, including `css`. Treeshaking only applies to `esm` so when importing `lib` it works fine. 

**Solution**
Setting `sideEffects: ["*.css"]` seems to solve the problem. 
This is an alternative solution to #55 and subset of #56.

**Reference**
https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free


khtruong (approved):

williaster:
wow that's tricky, nice fix 👍

PR #66, opened 2019-04-18
feat: improve line chart margin/axis and add buildquery

kristw:
🏆 Enhancements
* feat: can disable axis title
* feat: adjust margin according to axis title visibility
* feat: add type for line chart's formdata
* feat: add buildQuery for line chart

📜 Documentation
* docs: add more knobs to storybook

🐛 Bug Fix
* fix: fallback to default margin when margin is partially set

🏠 Internal
* chore: bump storybook addons version


packages/superset-ui-preset-chart-xy/src/Line/Line.tsx
```diff
@@ -22,9 +22,11 @@ import ChartLegend from '../components/ChartLegend';
 
 chartTheme.gridStyles.stroke = '#f1f3f5';
 
+const DEFAULT_MARGIN = { top: 20, right: 20, left: 20, bottom: 20 };
+
 const defaultProps = {
   className: '',
-  margin: { top: 20, right: 20, left: 20, bottom: 20 },
+  margin: DEFAULT_MARGIN,
```
schillerk, line 29: I don't think you need this since you're doing `margin: { ...DEFAULT_MARGIN, ...margin },` down below anyway

packages/superset-ui-preset-chart-xy/src/encodeable/AxisAgent.ts
```diff
@@ -54,8 +54,21 @@ export default class AxisAgent<Def extends ChannelDef<Output>, Output extends Va
     return this.format || this.channelEncoder.formatValue;
   }
 
+  hasTitle() {
+    return this.getTitle() !== undefined;
+  }
+
   getTitle() {
-    return this.config.title || this.channelEncoder.getTitle();
+    if (typeof this.config.title === 'undefined' || this.config.title === true) {
```
schillerk: Since it's an object property I *think* you can safely do `this.config.title === undefined`?

packages/superset-ui-preset-chart-xy/src/utils/XYChartLayout.tsx
```diff
@@ -154,7 +154,7 @@ export default class XYChartLayout {
 
     return axis && this.xLayout ? (
       <XAxis
-        label={axis.getTitle()}
+        label={axis.getTitle() || ''}
```
schillerk: Why not have getTitle return `''` instead of undefined, and then have `hasTitle` check against that instead of undefined?

Because we want different behavior for a user defined empty title?

schillerk (approved):
Woo thanks!

packages/superset-ui-preset-chart-xy/src/Line/Line.tsx
kristw: It is useful for deriving the type as well. (`typeof defaultProps`)

packages/superset-ui-preset-chart-xy/src/Line/buildQuery.ts
```diff
@@ -0,0 +1,6 @@
+import { buildQueryContext } from '@superset-ui/chart';
+import ChartFormData from './ChartFormData';
+
+export default function buildQuery(formData: ChartFormData) {
+  return buildQueryContext(formData);
```
conglei, line 5: Note that `buildQueryContext` cannot handle the time shift functions.

packages/superset-ui-preset-chart-xy/src/Line/transformProps.ts
```diff
@@ -4,13 +4,14 @@ import { ChartProps } from '@superset-ui/chart';
 
 export default function transformProps(chartProps: ChartProps) {
   const { width, height, formData, payload } = chartProps;
-  const { encoding } = formData;
+  const { encoding, margin } = formData;
   const { data } = payload;
 
   return {
     data,
```
conglei, line 11: Here, we need to implement `get_data` logic here.  `https://github.com/apache/incubator-superset/blob/master/superset/viz.py#L1222`

packages/superset-ui-preset-chart-xy/src/Line/buildQuery.ts
conglei: https://github.com/apache/incubator-superset/blob/master/superset/viz.py#L1194

packages/superset-ui-preset-chart-xy/src/encodeable/AxisAgent.ts
```diff
@@ -54,8 +54,21 @@ export default class AxisAgent<Def extends ChannelDef<Output>, Output extends Va
...
+      return this.channelEncoder.getTitle();
+    }
+
+    const { title } = this.config;
+    if (title === 'false' || title === false || title === '') {
```
conglei: it is a bit weird that we need to test `title === `false``. @kristw what is the reason behind it?

packages/superset-ui-preset-chart-xy/src/encodeable/AxisAgent.ts
```diff
@@ -54,8 +54,21 @@ export default class AxisAgent<Def extends ChannelDef<Output>, Output extends Va
...
+    return this.getTitle() !== undefined;
+  }
+
   getTitle() {
-    return this.config.title || this.channelEncoder.getTitle();
+    if (typeof this.config.title === 'undefined' || this.config.title === true) {
+      return this.channelEncoder.getTitle();
+    }
+
+    const { title } = this.config;
```
conglei, line 62: nit: we can move the Destructuring assignment right above line 62

packages/superset-ui-preset-chart-xy/src/utils/XYChartLayout.tsx
conglei: The logic here is bit complex. Can we make `axis.getTitle()` return `''` when there is no title, and for `hasTitle`, we can just test if `axis.getTitle() === ''`. I think it will make the code cleaner.

packages/superset-ui-preset-chart-xy/src/encodeable/AxisAgent.ts
kristw: agree. will remove this. storybook knobs didn't pass me boolean as expected.

packages/superset-ui-preset-chart-xy/src/Line/buildQuery.ts
kristw: It's ok. this will not handle time shift yet

packages/superset-ui-preset-chart-xy/src/utils/XYChartLayout.tsx
kristw: yep done

packages/superset-ui-preset-chart-xy/src/Line/transformProps.ts
kristw: If I don't implement anything, will it behave just like the word cloud query? I plan to get this pr in as a starter and will develop the buildQuery in a separate pr.

conglei (approved):
LGTM! Thanks for adding this :)

PR #68, opened 2019-04-23
fix: line chart does not handle temporal field correctly

kristw:
🐛 Bug Fix

The Line Chart is using `@data-ui/xy-chart` under the hood which takes care of scaling by itself at the moment. The previous code already encode the data from time into position before passing to `@data-ui/xy-chart` so it did not render correctly.


xtinec (approved):
LGTM 🎉

PR #71, opened 2019-04-23
feat: make line chart query operate in time series mode

kristw:
🐛 Bug Fix

To ensure the output is grouped by time for time series. 

@conglei

### DIFF
--- a/packages/superset-ui-preset-chart-xy/src/Line/buildQuery.ts
+++ b/packages/superset-ui-preset-chart-xy/src/Line/buildQuery.ts
@@ -2,5 +2,13 @@ import { buildQueryContext } from '@superset-ui/chart';
 import ChartFormData from './ChartFormData';
 
 export default function buildQuery(formData: ChartFormData) {
-  return buildQueryContext(formData);
+  const queryContext = buildQueryContext(formData);
+
+  // Enforce time-series mode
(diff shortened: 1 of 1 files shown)


packages/superset-ui-preset-chart-xy/src/Line/buildQuery.ts
```diff
@@ -2,5 +2,13 @@ import { buildQueryContext } from '@superset-ui/chart';
 import ChartFormData from './ChartFormData';
 
 export default function buildQuery(formData: ChartFormData) {
-  return buildQueryContext(formData);
+  const queryContext = buildQueryContext(formData);
+
+  // Enforce time-series mode
+  queryContext.queries.forEach(query => {
+    const q = query;
```
conglei, line 9: nit: Should we make a copy? (I don't think it is necessary imo).

conglei (approved):
LGTM :)

packages/superset-ui-preset-chart-xy/src/Line/buildQuery.ts
kristw: umm good point. but since nobody has access to the original one I think it is ok to modify instead of cloning the entire `queryContext`.

PR #72, opened 2019-04-24
feat: update data format returned from query api and add getGroupBys

kristw:
🐛 Bug Fix

- Update expected data format from `/api/v1/query` from 

```ts
type Dataset<T extends string = string> = {
  keys: T[];	
  values: Partial<PlainObject<T>>[];	
}
```

to

```ts
type Dataset<T extends string = string> = Partial<PlainObject<T>>[];
```

- Update corresponding code to no longer use `dataset.keys`.
- Add functions for getting `groupby` from encoding.


williaster (approved):
Nice! 😎👍

PR #74, opened 2019-04-26
fix: lazily create metadata

kristw:
🏆 Enhancements

* Delay creation of chart `metadata` which calls the translation function.


williaster (approved):
Is this an issue because the translation has to happen dynamically?

kristw:
There is an issue in the dataportal tests that crashes when `t()` is called before it is configured. That is super weird because it will only issue warning if not configured so I am not sure why it crashes.

However, by creating constant `metadata` and return. `t()` is called as soon as this file is imported. A little bit harder to organize import order to ensure `configure()` is called before `t()`. By delaying its creation until the plugin instance is instantiated. By that time the translation module should have been configured.

PR #82, opened 2019-05-07
fix: disable lazy import for line chart and box plot

kristw:
🏠 Internal

* Disable lazy `import` temporarily due to problem with downstream application. 
* Also fix broken build


williaster (approved):
sorry to go backward on this :( LGTM if the mock `loader` doesn't work 👍

kristw:
build broken due to changes in `eslint` `react/sort-comp` changes how it thinks things should be ordered -*-. just pushed a fix.

PR #106, opened 2019-05-20
feat(tablevis): add a new tablevis plugin

conglei:
💔 Breaking Changes

🏆 Enhancements
This pr is to rewrite current TableVis plugin using @airbnb/lunar DataTable component, which uses react-virtualizer to render the table. 

📜 Documentation

🐛 Bug Fix

🏠 Internal


packages/superset-ui-plugin-chart-table/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "url": "https://github.com/apache-superset/superset-ui-plugins/issues"
+  },
+  "homepage": "https://github.com/apache-superset/superset-ui-plugins#readme",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@airbnb/lunar": "^1.8.1",
+    "@airbnb/lunar-icons": "^1.1.1",
+    "core-js": "^3.0.1",
```
kristw: Is it `dependencies` or `devDependencies`? Could we pin it at root or it has to be here?

packages/superset-ui-plugin-chart-table/src/Table.tsx
```diff
@@ -0,0 +1,276 @@
...
+            <div style={boxStyle}>
+              <div style={boxContainerStyle}>
+                <div style={barStyle} />
+                <div style={{ zIndex: 10, marginLeft: 'auto' }}>{children}</div>
+              </div>
+            </div>
+          );
+        };
+
+        if (key[0] === '%') {
```
kristw: I also wonder if we can find better way to configure this than this hard coded rule

packages/superset-ui-plugin-chart-table/src/Table.tsx
```diff
@@ -0,0 +1,276 @@
...
+    } else {
+      newSelectedCells.add(cellHash);
+      onAddFilter(cell.key, [cell.value]);
+    }
+    this.setState({
+      selectedCells: newSelectedCells,
+    });
+  };
+
+  static getDerivedStateFromProps(props: TableProps, state: TableState) {
```
williaster: Should we condition this update on a filter change?

packages/superset-ui-plugin-chart-table/src/Table.tsx
```diff
@@ -0,0 +1,276 @@
...
+      }
+    }
+
+    const tsFormatter = getTimeFormatter(tableTimestampFormat);
+
+    const heightType = 'small';
+    const getRenderer = (column: ColumnType) => (props: RendererProps) => {
+      const { key } = props;
+      let value = props.row.rowData.data[key];
+      const cellHash = `${key}#${value}`;
```
williaster: Could define a method to get this to make sure it's the same here and in lifecycle methods

packages/superset-ui-plugin-chart-table/src/transformProps.ts
```diff
@@ -0,0 +1,124 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/* eslint-disable sort-keys */
+
+import { ChartProps, Metric, FormDataMetric } from '@superset-ui/chart';
+
+const DTTM_ALIAS = '__timestamp';
```
williaster, line 25: Could use this in the component file also?

williaster (commented):
Looking great!

packages/superset-ui-plugin-chart-table/src/Table.tsx
conglei: @williaster  I think this function should be always called whenever there is a prop change. Could you explain a bit more since I'm a bit confused 😆

conglei:
Note that currently table searching is not working properly. Will bump the version of lunar once https://github.com/airbnb/lunar/pull/70 gets merged.

cc @schillerk

packages/superset-ui-plugin-chart-table/src/Table.tsx
```diff
@@ -0,0 +1,206 @@
...
+        return content.indexOf(value) >= 0;
+      });
+      this.setState({
+        searchKeyword: value,
+        filteredRows,
+      });
+    }
+  };
+
+  static getDerivedStateFromProps: React.GetDerivedStateFromProps<
```
williaster: @conglei I just meant that this is called every time that props change.

So if `filters` have not changed between updates, you wouldn't need to update state right? (in that case you can return `null` from this func). If I'm misunderstanding you can ignore this :)

mistercrunch:
Oh just saw this https://github.com/airbnb/lunar, looks promising, would love to hear more about it :)

kristw:
Master build is back! Please rebase for storybook preview.

packages/superset-ui-plugin-chart-table/src/Table.tsx
conglei: Ah, gotcha, good catch! will revise the logic :)

williaster:
@conglei definitely ping me when you want another review on this

packages/superset-ui-plugin-chart-table/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@airbnb/lunar": "^1.13.0",
+    "@airbnb/lunar-icons": "^1.1.1",
+    "react": "^16.8.6"
+  },
+  "peerDependencies": {
+    "@superset-ui/chart": "^0.10.0",
```
kristw: You may want to bump these versions.

packages/superset-ui-plugin-chart-table/src/Table.tsx
```diff
@@ -0,0 +1,221 @@
...
+        };
+      }
+    });
+
+    return (
+      <>
+        {includeSearch && (
+          <div {...css(styles.searchBar)}>
+            <div {...css(styles.searchBox)}>
+              <Input
```
kristw, line 182: Does the `DataTable` now include built-in search box?

packages/superset-ui-plugins-demo/storybook/stories/plugin-chart-table/Stories.jsx
```diff
@@ -0,0 +1,110 @@
...
+import { SuperChart } from '@superset-ui/chart';
+import dataLegacy from './dataLegacy';
+import data from './data';
+
+export default [
+  {
+    renderStory: () => (
+      <SuperChart
+        chartType="table2-legacy"
+        chartProps={{
```
kristw: Check the other examples are updated to use new `SuperChart` api that spread out the fields as top-level props instead of nesting under `chartProps`

packages/superset-ui-plugin-chart-table/src/Table.tsx
conglei: There is no built-in search box for `<DataTable/>`, and in any case, users need to implement the search logic by themself, the example here is https://github.com/airbnb/lunar/blob/master/packages/core/src/components/DataTable.story.tsx#L121.  

However, using it will prevent us from showing the message like `show ** out of ** row`, and the benefit of using `filterData` is to keep the status of `expended rows`, which we don't really use here.  

So I think we should just use the logic written here.

packages/superset-ui-plugin-chart-table/src/buildQuery.ts
```diff
@@ -0,0 +1,35 @@
...
+    const sortby = formData.timeseries_limit_metric;
+    if (formData.all_columns && formData.all_columns.length > 0) {
+      columns = [...formData.all_columns];
+      const orderByColumns = formData.order_by_cols || [];
+      orderByColumns.forEach(columnOrder => {
+        const parsedColumnOrder: [ChartFormDataMetric, boolean] = JSON.parse(columnOrder);
+        orderby.push([convertMetric(parsedColumnOrder[0]), parsedColumnOrder[1]]);
+      });
+      groupby = [];
+    } else if (sortby) {
```
kristw, line 21: Is it `else if` or `if`?

packages/superset-ui-plugin-chart-table/src/processColumns.ts
```diff
@@ -0,0 +1,92 @@
...
+    if (metrics.indexOf(key) >= 0) {
+      formatFunction = getNumberFormatter(formatString);
+      type = 'metric';
+      extraField.maxValue = maxes[key];
+      extraField.minValue = mins[key];
+    }
+
+    // Handle verbose names for percents
+    if (!label) {
+      if (key[0] === '%') {
```
kristw: Add `key.length > 0`?

packages/superset-ui-plugin-chart-table/src/processData.ts
```diff
@@ -0,0 +1,47 @@
...
+  let processedData: {
+    data: PlainObject;
+  }[] = records.map((row: PlainObject) => ({
+    data: row,
+  }));
+
+  if (sortByKey) {
+    processedData = processedData.sort((a, b) => {
+      const delta = a.data[sortByKey] - b.data[sortByKey];
+      if (orderDesc) {
```
kristw: can be slightly faster if do the `orderDesc` check outside of sort fn.

```
processedData = processData.sort(orderDesc
  ? (a, b) => b.data[sortByKey] - a.data[sortByKey] 
  : (a, b) => a.data[sortByKey] - b.data[sortByKey]);
```

can also sort raw records first before mapping to only map once.

```
const sortedData = records.slice().sort(orderDesc
  ? (a, b) => b[sortByKey] - a[sortByKey] 
  : (a, b) => a[sortByKey] - b[sortByKey]));
return sortedData.map((metrics.indexOf(sortByKey) < 0) 
  ? row => {
    const data = { ...row };
    delete data[sortByKey];
    return { data };
  }) 
  : row => ({ data: row })
);
```

conglei:
@kristw  addressed the issues you mentioned :)

packages/superset-ui-plugin-chart-table/src/processMetrics.ts
```diff
@@ -0,0 +1,22 @@
...
+}) {
+  const processedMetrics = (metrics || []).map(m => (m as AdhocMetric).label || (m as string));
+
+  const processedPercentMetrics = (percentMetrics || [])
+    .map(m => (m as AdhocMetric).label || (m as string))
+    .map(m => `%${m}`);
+
+  return processedMetrics
+    .concat(processedPercentMetrics)
+    .filter(m => typeof records[0][m as string] === 'number');
```
kristw: would `(m: string) => ...` avoid the casting `m as string`

packages/superset-ui-plugin-chart-table/src/processMetrics.ts
```diff
@@ -0,0 +1,22 @@
...
+export default function processMetrics({
+  metrics,
+  percentMetrics,
+  records,
+}: {
+  metrics: ChartFormDataMetric[];
+  percentMetrics: ChartFormDataMetric[];
+  records: PlainObject[];
+}) {
+  const processedMetrics = (metrics || []).map(m => (m as AdhocMetric).label || (m as string));
```
kristw, line 13: Can declare default param in the function signature and let `metrics` and `percentMetrics` be optional. (Although this won't work with `metrics` or `percentMetrics` being `null`)

```
export default function processMetrics({
  metrics = [],
  percentMetrics = [],
  records,
}: {
  metrics?: ChartFormDataMetric[];
  percentMetrics?: ChartFormDataMetric[];
  records: PlainObject[];
})
```

packages/superset-ui-plugin-chart-table/src/renderer.tsx
```diff
@@ -0,0 +1,115 @@
...
+      const barStyle: CSSProperties = {
+        background: color,
+        borderRadius: 3,
+        height: HEIGHT_TO_PX[heightType] / 2 + 4,
+        left: `${left}%`,
+        position: 'absolute',
+        width: `${width}%`,
+      };
+
+      const numberStyle: CSSProperties = {
```
kristw: This constant can be moved outside of function.

kristw:
A few minor things but overall lgtm

packages/superset-ui-plugin-chart-table/src/buildQuery.ts
conglei: It is `else if` according to the logic here: https://github.com/apache/incubator-superset/blob/master/superset/viz.py#L536

packages/superset-ui-plugin-chart-table/src/processMetrics.ts
conglei: for current control panel logic, I think the metrics and percentMetrics will be always a valid array. If there is no metric / percentMetrics, they will be just empty arrays.  For safety check, I added this logic.

kristw:
mm. broken build.

conglei:
@kristw build passed :)!

kristw (approved):
Please merge when you are ready.

PR #127, opened 2019-06-19
 docs: update storybook to use new SuperChart props and convert to TS

kristw:
📜 Documentation

* Update examples in storybook to use new `SuperChart` props from version `0.11.11` that no longer nest properties under `chartProps`.
* Change example file format to typescript when applicable.


etr2460 (approved):
lgtm, everything seems reasonable

PR #132, opened 2019-06-26
fix: heatmap errors with null data

michellethomas:
🐛 Bug Fix

Heatmap throws a js error when x or y data contains nulls. Previously we were replacing the nulls with strings, but with John's recent change in superset we are keeping them as nulls and need to handle them on the frontend correctly. To test the fix I added another item to storybook that contains nulls. 

@kristw @williaster @etr2460


packages/superset-ui-legacy-plugin-chart-heatmap/src/Heatmap.js
```diff
@@ -109,8 +109,8 @@ function Heatmap(element, props) {
 
     for (let i = 0; i < records.length; i++) {
       const datum = records[i];
-      longestX = Math.max(longestX, datum.x.toString().length || 1);
-      longestY = Math.max(longestY, datum.y.toString().length || 1);
+      longestX = Math.max(longestX, datum.x ? datum.x.toString().length || 1 : 1);
```
etr2460: not sure which is more readable, but this statement would only require specifying 1 once:
`(datum.x && datum.x.toString().length) || 1`

etr2460 (approved):
one comment, otherwise lgtm

packages/superset-ui-legacy-plugin-chart-heatmap/src/Heatmap.js
michellethomas: I like that better thanks

PR #134, opened 2019-07-02
fix: fixing tooltip for expanded area chart

michellethomas:
🐛 Bug Fix

[This PR](https://github.com/apache-superset/superset-ui-plugins/pull/27) added functionality for more data in the area chart rich tooltip, but those calculations don't make sense for the expanded chart. The expanded chart tooltip already shows % of total values and doesn't have a total line item. `generateAreaChartTooltipContent` shouldn't be called for `expand` stacked style.

image

@graceguo-supercat @etr2460

### DIFF
--- a/packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
+++ b/packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
@@ -559,7 +559,7 @@ function nvd3Vis(element, props) {
         chart.interactiveLayer.tooltip.contentGenerator(d =>
           generateRichLineTooltipContent(d, smartDateVerboseFormatter, yAxisFormatter),
         );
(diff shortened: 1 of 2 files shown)


etr2460 (commented):
How does this look when you click one of the items in the legend? In the current behavior, I think the tooltip has additional information when you unselect series from the chart

michellethomas:
expanded:
image

stacked:
image

It's not super necessary to show the 100% on the stacked chart but it's not incorrect so it seems ok.

etr2460:
Sorry, I meant when a subset of series are selected, like this:
image

The percents there provide additional information as the fraction of the selected series that each individual series makes up.

michellethomas:
Oh I see, it just adjusts the %:
image

etr2460 (approved):
lgtm

PR #146, opened 2019-07-24
build: automate build & launch of storybook

andreaslordos:
🏆 Enhancements
- Added `build.py` in the `/packages/superset-ui-plugins-demo` directory
	- Running `build.py` will build & start storybook automatically
 	- No need to run 4 separate commands to do so
 

📜 Documentation
 - Modified README.md to reflect this change and provide instructions on using `build.py`

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,12 @@ cd packages/superset-ui-plugins-demo
 yarn storybook
 ```
 
+Alternatively, you can use the following command while in `packages/superset-ui-plugins-demo`
+
+```sh
+sudo python build.py
+```
+
 ### Committing
 
 This repository follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0-beta.3/) guideline for commit messages and has a `commitlint` hook which will require you to have the valid commit message before committing. You can use `npm run commit` to help you create a commit message.
--- a/README.md
+++ b/README.md
@@ -109,10 +109,10 @@ cd packages/superset-ui-plugins-demo
 yarn storybook
 ```
 
-Alternatively, you can use the following command while in `packages/superset-ui-plugins-demo`
+Alternatively, you can demo your changes by using the following command while in `packages/superset-ui-plugins-demo`:
 
 ```sh
-sudo python build.py
+npm run build
 ```
 
 ### Committing
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ You can demo your changes by running the storybook demo locally with the followi
 yarn install
 yarn build
 cd packages/superset-ui-plugins-demo
-yarn storybook
+yarn storybook:run
 ```
 
 Alternatively, you can demo your changes by using the following command while in `packages/superset-ui-plugins-demo`:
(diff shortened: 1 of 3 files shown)


etr2460 (changes requested):
cool idea! Although I think the more standard way would be to add a command in the demo `package.json` file. Then you could just use bash as opposed to python to do `cd ../../ && yarn build && cd - && yarn storybook`

kristw:
You also only need to build once. The subsequent storybook calls should not need to rebuild

andreaslordos:
> You also only need to build once. The subsequent storybook calls should not need to rebuild

@kristw I believe if you edit the code of a plugin/add your own, you will need to build Storybook again for the changes to take effect.

kristw:
@andreaslordos You can deep import from `@superset-ui/xxx/src/yyy` instead of `@superset-ui/xxx` (which will point to the built files) during development.

andreaslordos:
> cool idea! Although I think the more standard way would be to add a command in the demo `package.json` file. Then you could just use bash as opposed to python to do `cd ../../ && yarn build && cd - && yarn storybook`

@etr2460 thank you for the suggestion! I have modified my PR accordingly

packages/superset-ui-plugins-demo/package.json
```diff
@@ -9,7 +9,8 @@
     "demo:build": "build-storybook -o _gh-pages",
     "demo:publish": "gh-pages -d _gh-pages",
     "deploy-demo": "npm run demo:clean && npm run demo:build && npm run demo:publish && npm run demo:clean",
-    "storybook": "start-storybook -p 9001"
+    "storybook": "start-storybook -p 9001",
+    "build": "cd ../../ && yarn build && cd - && yarn storybook"
```
etr2460: Much better! Maybe one final nit, I'd name this something like `build-and-run` or maybe follow the example with `demo`, naming the old `storybook` command `storybook:run` and this one just `storybook`

etr2460 (commented):
Make sure you address the commit lint issues too (format the commit title properly)

packages/superset-ui-plugins-demo/package.json
andreaslordos: Addressed commit lint issues (I think) & renamed the script name to "storybook" - thank you for the suggestions!

README.md
```diff
@@ -106,7 +106,13 @@ You can demo your changes by running the storybook demo locally with the followi
...
 yarn build
 cd packages/superset-ui-plugins-demo
-yarn storybook
+yarn storybook:run
+```
+
+Alternatively, you can demo your changes by using the following command while in `packages/superset-ui-plugins-demo`:
+
+```sh
+npm run storybook
```
etr2460: one final nit, this should be `yarn storybook`

etr2460 (changes requested):
almost there!

README.md
andreaslordos: @etr2460 fixed

andreaslordos:
Requesting review @etr2460

etr2460 (approved):
lgtm, thanks for the contribution!

PR #148, opened 2019-07-24
fix: single y axis bounds

etr2460:
🐛 Bug Fix

Further fixes to the custom y axis bounds for nvd3 charts. We were only respecting custom y axis bounds if they were both set; this allows the user to only set 1 custom y axis bound. The other one will be automatically calculated based on what type of chart this is and what the max/min data values are.

Note that this also special cases expanded area and stream area charts from having custom y axis bounds. These charts don't make much sense without the default bounds, so I've disabled them.

Screenshots:
Screen Shot 2019-07-24 at 4 36 43 PM
Screen Shot 2019-07-24 at 4 36 52 PM
bounds 0 - 3 billion
Screen Shot 2019-07-24 at 4 37 15 PM

bounds 1 billion - null
Screen Shot 2019-07-24 at 4 37 24 PM
Screen Shot 2019-07-24 at 4 37 31 PM

ignore custom bounds
Screen Shot 2019-07-24 at 4 37 38 PM

bar charts without bounds still work
Screen Shot 2019-07-24 at 4 37 47 PM
Screen Shot 2019-07-24 at 4 37 53 PM

reviewers: @khtruong @kristw @williaster @graceguo-supercat

### DIFF
--- a/packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
+++ b/packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
@@ -36,6 +36,8 @@ import isTruthy from './utils/isTruthy';
 import {
   cleanColorInput,
   computeBarChartWidth,
+  computeYDomain,
+  computeStackedYDomain,
   drawBarValues,
   generateBubbleTooltipContent,
   generateMultiLineTooltipContent,
@@ -440,7 +442,7 @@ function nvd3Vis(element, props) {
 
     if (showBarValue) {
       drawBarValues(svg, data, isBarStacked, yAxisFormat);
-      chart.dispatch.on('stateChange', () => {
+      chart.dispatch.on('stateChange.drawBarValues', () => {
         drawBarValues(svg, data, isBarStacked, yAxisFormat);
       });
     }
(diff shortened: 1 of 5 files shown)


packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
```diff
@@ -607,16 +609,54 @@ function nvd3Vis(element, props) {
...
+          // Because there are custom bounds, we need to override them back to 0%-100% since this
+          // is an expanded area chart
+          chart.yDomain([0, 1]);
+        } else if (
+          (hasCustomMin || hasCustomMax) &&
+          vizType === 'area' &&
+          chart.style() === 'stream'
+        ) {
+          // Because there are custom bounds, we need to override them back to the domain of the
+          // chart since this is a stream area chart
```
williaster: domain of the `data`?

packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
```diff
@@ -607,16 +609,54 @@ function nvd3Vis(element, props) {
...
+          chart.yDomain([min, max]);
+          chart.clipEdge(true);
+        }
       }
-    }
+    };
+    applyYAxisBounds();
+
+    // Also reapply on each state change to account for enabled/disabled series
+    chart.dispatch.on('stateChange.applyYAxisBounds', () => {
```
williaster: can we remove the extra arrow func here? 

`chart.dispatch.on('stateChange.applyYAxisBounds', applyYAxisBounds)`

williaster (approved):
This LGTM. I agree about disabling custom bounds for expanded (0-100%) area charts, and I also agree with using the domain for stream area charts 👍

packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
etr2460: Done.

packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
etr2460: good catch, done!

etr2460:
tests added too

williaster:
🔥

PR #159, opened 2019-07-29
fix: nvd3 charts break on stateChange dispatch

etr2460:
🐛 Bug Fix
The change in https://github.com/apache-superset/superset-ui-plugins/pull/148 broke charts that don't have a `stateChange` dispatch event like the box plot. This fixes the issue.

Tested by opening all the charts (including box plot) in storybook and seeing them render

@kristw

### DIFF
--- a/packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
+++ b/packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
@@ -654,7 +654,9 @@ function nvd3Vis(element, props) {
     applyYAxisBounds();
 
     // Also reapply on each state change to account for enabled/disabled series
-    chart.dispatch.on('stateChange.applyYAxisBounds', applyYAxisBounds);
+    if (chart.dispatch && chart.dispatch.stateChange) {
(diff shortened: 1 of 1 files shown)


packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
```diff
@@ -654,7 +654,9 @@ function nvd3Vis(element, props) {
     applyYAxisBounds();
 
     // Also reapply on each state change to account for enabled/disabled series
-    chart.dispatch.on('stateChange.applyYAxisBounds', applyYAxisBounds);
+    if (chart.dispatch && chart.dispatch['stateChange']) {
```
kristw: `chart.dispatch.stateChange`?

kristw (approved):

packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
etr2460: oops, duh. fixed!

PR #160, opened 2019-07-30
build: speed up storybook build

kristw:
🏠 Internal

* Avoid parsing large libraries and reduce optimization aggressiveness to trade for speed and hopefully reduce netlify timeout.
* Disable storybooks for MapBox and CountryMap which are not actively developed and included large dependencies.


etr2460:
looks like the preview doesn't work :/ some error in the es6-shim

kristw:
@etr2460 I disabled stories of two components (CountryMap and MapBox) and it seems to build for now. Could be coupling with `netlify` cluster suddenly has more availability.

etr2460 (approved):
lgtm. Maybe add a comment for why we commented out those files?

PR #162, opened 2019-07-30
feat: file skeleton necessary for icicle chart visualization

cguan7:
🏆 Enhancements

Initial PR to layout file skeleton to implement Icicle Chart


packages/superset-ui-plugin-chart-icicle-event/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+```js
+import IcicleEventVizPlugin from '@superset-ui/plugin-chart-icicle-event';
+
+new IcicleEventVizPlugin()
+  .configure({ key: 'icicle' })
+  .register();
+```
+
+Then use it via `SuperChart`. See [storybook](https://apache-superset.github.io/superset-ui-plugins/?selectedKind=plugin-chart-word-cloud) for more details.
```
khtruong: Might want to update the link in a later PR since IcicleEventChart is not available yet.

packages/superset-ui-plugin-chart-icicle-event/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@superset-ui/plugin-chart-icicle",
```
khtruong: Might want to re-generate this with icicle-event

packages/superset-ui-plugin-chart-icicle-event/src/IcicleEventChart.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { Component, createRef } from 'react';
```
khtruong, line 19: Add file headers. Comment applies to the other files. This should make one of the build pre-checks pass.

cguan7: Added the file headers to all of the files but the checks are failing due to a timeout

I believe this PR is related and will be a fix for the Netlify checks: 
https://github.com/apache-superset/superset-ui-plugins/pull/160

khtruong (approved):

PR #165, opened 2019-08-01
feat: add basic functionality for icicle chart to display static data

cguan7:
🏆 Enhancements
adds functionality for the icicle chart to have data passed in and display (static, no interactions
yet)

### DIFF
--- a/packages/superset-ui-plugin-chart-icicle-event/src/IcicleEventChart.tsx
+++ b/packages/superset-ui-plugin-chart-icicle-event/src/IcicleEventChart.tsx
@@ -18,7 +18,9 @@
  */
 import React, { Component, createRef } from 'react';
 import { HierarchyRectangularNode } from 'd3-hierarchy';
+import { select as d3Select } from 'd3-selection';
 import { IcicleEventNode } from '../types/IcicleEventNode';
+import { x0, y0, rectWidth, rectHeight } from './utils/RenderedIcicleAccessors';
 
 interface Props {
   className?: string;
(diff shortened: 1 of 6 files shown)


williaster:
do you have any screenshots?

packages/superset-ui-plugin-chart-icicle-event/src/IcicleEventChart.tsx
```diff
@@ -50,10 +79,82 @@ export default class IcicleEventChart extends Component<Props> {
...
+    const { boxMargin, color, contentRenderer, isVertical, width, height, rounding } = this.props;
+
+    const root = this.props.d3TreeRoot;
+
+    // Clear all elements and redraw the Icicle Chart
+    d3Select(this.chartRef.current)
+      .selectAll('*')
+      .remove();
+
+    const svg = d3Select(this.chartRef.current)
```
williaster, line 109: I'm curious why you aren't using a library like [`@vx` ](https://vx-demo.now.sh/gallery) that plays better with `react` and renders visualizations/utilizes `d3` more declaratively.

cguan7:
Yes, I have some some screenshots! This is what it looks like currently in Storybook with the intention that it is not the final look. Since this PR was already getting a little big, I wanted to make it in a separate one:

image

Additionally, the README for this package has a screenshot of a prototype that is slightly outdated (we're still going through design, but have made some changes to spacing between icicles, etc.)

packages/superset-ui-plugin-chart-icicle-event/src/IcicleEventChart.tsx
cguan7: I have no good answer for that other than being not very familiar with vx... I recall seeing it in the `event-flow` viz if I remember correctly but didn't look too much into it. Will definitely look more into that library tomorrow in the morning!

packages/superset-ui-plugin-chart-icicle-event/src/IcicleEventChart.tsx
```diff
@@ -29,14 +31,41 @@ interface Props {
     y: number;
   };
   color: (name: string) => string;
-  contentRenderer: () => void;
+  contentRenderer: (...args: any[]) => void;
```
khtruong: Can we more more explicit on the type?

packages/superset-ui-plugin-chart-icicle-event/src/IcicleEventChart.tsx
khtruong: Very good point! The visualization will have some animation as well and it doesn't look like vx supports it. Although we could use other react animation libraries with vx. I think given Chris's deadlines, I suggest that this is something we look into later on. I am not a big fan of how unreadable d3 either but fortunately, I don't think there is much more than what is already in this PR.

packages/superset-ui-plugin-chart-icicle-event/src/IcicleEventChart.tsx
cguan7: Updated to include the types, `BaseType` is documented here:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/d3-selection/index.d.ts#L21

khtruong (approved):

PR #167, opened 2019-08-05
fix: only remove tooltips relating to a single vis

etr2460:
🐛 Bug Fix

When rerendering charts, we were removing all tooltips for every chart from the dom instead of just the one owned by the rerendering chart. This would break pages with multiple nvd3 charts on them (basically all superset dashboards). It can be easily reproed with dashboards with multiple tabs with the following steps:
1. hover over a chart on tab 1
2. switch to tab 2
3. resize the window, then resize it back to the original size
4. switch back to tab 1
5. see no more tooltips on the previously hovered chart. other charts will add new tooltips though

I fixed this issue by uniquely identifying charts by the passed in id and applying this id to the tooltips as well. When rerendering or unmounting charts now, if an id was passed in then we only remove the tooltip belonging to that chart. If no ids are passed, then we retain the current behavior.

### DIFF
--- a/packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
+++ b/packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
@@ -43,12 +43,14 @@ import {
   generateMultiLineTooltipContent,
   generateRichLineTooltipContent,
   generateTimePivotTooltip,
+  generateTooltipClassName,
   generateAreaChartTooltipContent,
   getMaxLabelSize,
   getTimeOrNumberFormatter,
   hideTooltips,
   tipFactory,
   tryNumify,
+  removeTooltip,
   setAxisShowMaxMin,
   stringifyTimeRange,
   wrapTooltip,
@@ -254,6 +256,10 @@ function nvd3Vis(element, props) {
   const container = element;
   container.innerHTML = '';
   const activeAnnotationLayers = annotationLayers.filter(layer => layer.show);
+  const chartId =
+    container.parentElement && container.parentElement.id !== ''
+      ? container.parentElement.id
+      : null;
 
   let chart;
   let width = maxWidth;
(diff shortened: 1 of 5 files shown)


packages/superset-ui-legacy-preset-chart-nvd3/src/ReactNVD3.js
```diff
@@ -18,10 +18,15 @@
  */
 import { reactify } from '@superset-ui/chart';
 import Component from './NVD3Vis';
-import { hideTooltips } from './utils';
+import { hideTooltips, removeTooltip } from './utils';
 
 function componentWillUnmount() {
-  hideTooltips(true);
+  const { id } = this.props; // eslint-disable-line babel/no-invalid-this
+  if (id) {
```
kristw: Can `id` be `0` ?

packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
```diff
@@ -254,6 +256,7 @@ function nvd3Vis(element, props) {
   const container = element;
   container.innerHTML = '';
   const activeAnnotationLayers = annotationLayers.filter(layer => layer.show);
+  const chartId = container.parentElement.id || null;
```
kristw: perhaps also check if `parentElement` is not null.
Also, do we want to prevent accidents if `id` is `0` which is falsy?

kristw (approved):
Overall LGTM, have minor comments on some pessimistic edge cases.

packages/superset-ui-legacy-preset-chart-nvd3/src/ReactNVD3.js
etr2460: updated to `id != null`

packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
etr2460: updated

PR #183, opened 2019-08-25
feat: upgrade @superset-ui to v0.12

kristw:
## 💔 Breaking Changes:

Many changes to the plugins are required to make them compatible with `@superset-ui/xxx` `v0.12`. After the changes in this PR, these plugin packages are no longer backward compatible with `@superset-ui/xxx` `v0.10` or `v0.11`, therefore worth calling out as breaking changes.

### 2. Update `transformProps.(js|ts)`

Several fields in `chartProps` are changed.

* rename `payload` to `queryData`
* rename `filters` to `initialValues`
* unnest `onAddFilter`, `onError`, `setControlValue`, `setTooltip` from `hooks`

### 3. Update types imported from `@superset-ui/chart`

* Several types are no longer in `@superset-ui/chart` and moved into the new package `@superset-ui/query` in `v0.12`
* A few types are also renamed.

### 4. Update storybooks and READMEs

Update `<SuperChart>` props

## Test

I have verified all stories in storybook and they work like before. 
Note: The legacy `table` and `pivot_table` examples are not working due to wrong webpack configuration which I will address in another PR.


packages/superset-ui-legacy-preset-chart-nvd3/src/transformProps.js
```diff
@@ -20,17 +20,12 @@
 import isTruthy from './utils/isTruthy';
 import { formatLabel } from './utils';
 
+function NOOP() {}
```
etr2460: is this preferred to `const NOOP = () => {};`? I find defining a function like this as a constant at the top of the file a bit more readable

etr2460 (approved):
one question about `NOOP`, otherwise lgtm!

packages/superset-ui-legacy-preset-chart-nvd3/src/transformProps.js
kristw: sure! can do.

PR #189, opened 2019-09-04
fix(tablevis): Set proper width for each column

conglei:
💔 Breaking Changes

🏆 Enhancements

📜 Documentation

🐛 Bug Fix
This PR is to fix the issue of the readability of table vis when there are too many columns in the table. 

image

Before rendering the table, we will estimate the width for each column according to the max string length (max width is set to be 500px). 

🏠 Internal

### DIFF
--- a/packages/superset-ui-plugin-chart-table/src/Table.tsx
+++ b/packages/superset-ui-plugin-chart-table/src/Table.tsx
@@ -9,6 +9,7 @@ import { getRenderer, ColumnType, heightType, Cell } from './renderer';
 type Props = {
   data: ParentRow[];
   height: number;
+  width: number;
   alignPositiveNegative?: boolean;
   colorPositiveNegative?: boolean;
   columns: ColumnType[];
@@ -33,6 +34,10 @@ const defaultProps = {
 };
 
 const SEARCH_BAR_HEIGHT = 40;
+
+const CHAR_WIDTH = 10;
+
+const MAX_COLUMN_WIDTH = 500;
 
 export type TableProps = Props & Readonly<typeof defaultProps>;
 
@@ -148,6 +153,7 @@ class TableVis extends React.PureComponent<InternalTableProps, TableState> {
       alignPositiveNegative,
       colorPositiveNegative,
       height,
+      width,
       tableFilter,
       styles,
       includeSearch,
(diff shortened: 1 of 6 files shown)


packages/superset-ui-plugins-demo/storybook/stories/plugin-chart-table/bigData.js
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable no-unused-vars */
+/* eslint-disable no-magic-numbers */
+/* eslint-disable sort-keys */
+const longString = `The quick brown fox jumps over the lazy dog`;
```
kristw: perhaps make these constant names `ALL_CAPS`

packages/superset-ui-plugin-chart-table/src/Table.tsx
```diff
@@ -176,6 +182,18 @@ class TableVis extends React.PureComponent<InternalTableProps, TableState> {
       }
     });
 
+    const keys = dataToRender && dataToRender.length > 0 ? Object.keys(dataToRender[0].data) : [];
+    let calculatedWidth = 0;
+    keys.forEach(key => {
+      const maxLength = Math.max(...data.map(d => String(d.data[key]).length), key.length);
+      columnMetadata[key] = {
+        maxWidth: MAX_COLUMN_WIDTH,
+        width: maxLength * CHAR_WIDTH,
```
kristw, line 191: Can use this cheap estimate of `CHAR_WIDTH` or 

```ts
import { getMultipleTextDimensions } from `@superset-ui/dimension`;

// can also specify text style
const dimensions = getMultipleTextDimensions({ texts: data.map(d => String(d.data[key]).concat([key]) });
const columnWidth = Math.max(...dimensions.map(dim => dim.width));
```

kristw: It will perform not as good as the heuristics but should give more accurate column width.

conglei: Will try this method :)

conglei: After trying out this method, we decided to use `CHAR_WIDTH` for faster computation, and later when we move the logic to `<DataTable/>`, we can re-evaluate the approach.

kristw (approved):

PR #199, opened 2019-09-12
feat(datatable): render html correctly

conglei:
💔 Breaking Changes

🏆 Enhancements
The pr is to add the capability to render the html correctly.  https://github.com/milesj/interweave is used to render the html.

📜 Documentation

🐛 Bug Fix

🏠 Internal

### DIFF
--- a/packages/superset-ui-plugin-chart-table/src/Table.tsx
+++ b/packages/superset-ui-plugin-chart-table/src/Table.tsx
@@ -4,6 +4,7 @@ import Text from '@airbnb/lunar/lib/components/Text';
 import Input from '@airbnb/lunar/lib/components/Input';
 import withStyles, { WithStylesProps } from '@airbnb/lunar/lib/composers/withStyles';
 import { Renderers, ParentRow, ColumnMetadata } from '@airbnb/lunar/lib/components/DataTable/types';
+import dompurify from 'dompurify';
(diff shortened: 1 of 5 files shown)


packages/superset-ui-plugin-chart-table/src/Table.tsx
```diff
@@ -58,6 +58,17 @@ function getCellHash(cell: Cell) {
   return `${cell.key}#${cell.value}`;
 }
 
+function getText(value: string | number) {
+  if (typeof value === 'string') {
+    const span = document.createElement('span');
+    span.innerHTML = value;
```
kristw: I would feel better if we sanitize `value` first. 
https://www.npmjs.com/package/sanitize-html

Plugging value into it directly seems like a security concern.

kristw: Or there might be a way to use part of interweave to do it.
https://github.com/milesj/interweave/blob/master/packages/core/src/Parser.ts

conglei: In the end, I chose dompurify to do this, since for the legacy tablevis, this library is used, and this library is also the smallest among several sanitization libraries.

kristw (approved):

Issue #125, opened 2019-06-11
Table filter not working


gbrian:
**Describe the bug**
Table filter does not remove filters and dashboards ends up storing the filters on metadata.

**To Reproduce**
1. Add a Table slice to dashboard with "Emit Filter" option enabled
2. Click on row (dashboard is filtered)
3. Click again same row (nothing happens)
4. Dashboard keeps filter forever

**Expected behavior**
Add/Remove filter

**Screenshots**

**Environment (please complete the following information):**
 - superset-ui version: latest
 - Node version: `node -v`
 - npm version: `npm -v`
 
**Additional context**
onRemoveFilter is never passed to the Table control. `transformProps`does not map any removeFilter option

PR #206, opened 2019-09-22
fix(#125): emit remove filter

cmyip:
🐛 Bug Fix
Fixes issue #125 , links to apache/incubator-superset#8273


kristw:
Have you verified that this works as intended? Could you add some explanation of the behaviors before/after?

PR #240, opened 2019-10-24
refactor: convert word cloud to full react

kristw:
🏆 Enhancements

* non-legacy `WordCloud` is converted to native react. 

📜 Documentation

* Add storybook

image

@conglei @khtruong


packages/superset-ui-plugin-chart-word-cloud/src/transformProps.ts
```diff
@@ -13,14 +13,14 @@ function transformData(data: ChartProps['queryData'][], formData: ChartProps['fo
...
-  const { colorScheme, rotation, sizeTo, sizeFrom } = formData;
+  const { colorScheme, rotation, sizeTo } = formData;
 
   return {
     colorScheme,
     data: transformData(queryData.data, formData),
     height,
     rotation,
-    sizeRange: [sizeFrom, sizeTo],
+    sizeRange: [0, sizeTo],
```
conglei, line 23: Why do we need to remove `sizeFrom`?

packages/superset-ui-plugin-chart-word-cloud/src/WordCloud.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  componentDidUpdate() {
+    this.update();
+  }
+
+  componentWillUnmount() {
+    this.isMounted = false;
+  }
+
+  update() {
```
conglei, line 67: From the story book, it seems that this method is running forever and in the end there is some console log error.

packages/superset-ui-plugin-chart-word-cloud/src/WordCloud.tsx
```diff
@@ -0,0 +1,104 @@
...
+        }
+      })
+      .start();
+  }
+
+  render() {
+    const { width, height, colorScheme } = this.props;
+    const { words } = this.state;
+
+    const colorFn = CategoricalColorNamespace.getScale(colorScheme);
```
conglei, line 94: Do we need the color for the word? since there is no semantic meaning of the color, and to me, it is pretty distracting.

packages/superset-ui-plugin-chart-word-cloud/src/WordCloud.tsx
kristw: Oh... Will check

packages/superset-ui-plugin-chart-word-cloud/src/WordCloud.tsx
kristw: I am trying to match current behavior.

packages/superset-ui-plugin-chart-word-cloud/src/transformProps.ts
kristw: I think size should start from `0`

kristw:
comments addressed

packages/superset-ui-plugin-chart-word-cloud/src/transformProps.ts
conglei: if it starts from '0', does it mean for the least frequent word, its font size will be 0?  We may need to set a better lower bound.  Also, does it mean we want to also deprecate it from wordcloud controls?

kristw: domain will be `[0, max(frequency)]`. Yes, should deprecate the control

PR #241, opened 2019-10-28
fix(table): fixed performance issue

conglei:
💔 Breaking Changes

🏆 Enhancements

📜 Documentation

🐛 Bug Fix

This PR is to fix the performance issue for calculating the text width by adding cacheing logic. 

@kristw 
🏠 Internal

### DIFF
--- a/packages/superset-ui-plugin-chart-table/src/Table.tsx
+++ b/packages/superset-ui-plugin-chart-table/src/Table.tsx
@@ -80,15 +80,8 @@ type columnWidthMetaDataType = {
 };
 
 class TableVis extends React.PureComponent<InternalTableProps, TableState> {
-  columnWidthSelector = createSelector<
-    ParentRow[],
-    ParentRow[],
-    {
-      columnWidthMetaData: columnWidthMetaDataType;
-      totalWidth: number;
-    }
-  >(
-    data => data,
+  columnWidthSelector = createSelector(
+    (data: ParentRow[]) => data,
     data => {
       const keys = data && data.length > 0 ? Object.keys(data[0].data) : [];
       let totalWidth = 0;
--- a/packages/superset-ui-plugin-chart-table/src/Table.tsx
+++ b/packages/superset-ui-plugin-chart-table/src/Table.tsx
@@ -5,6 +5,7 @@ import Input from '@airbnb/lunar/lib/components/Input';
 import withStyles, { WithStylesProps } from '@airbnb/lunar/lib/composers/withStyles';
 import { Renderers, ParentRow, ColumnMetadata } from '@airbnb/lunar/lib/components/DataTable/types';
 import dompurify from 'dompurify';
+import { createSelector } from 'reselect';
 import { getRenderer, ColumnType, Cell } from './renderer';
 
 type Props = {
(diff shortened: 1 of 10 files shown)


packages/superset-ui-plugin-chart-table/src/renderer.tsx
```diff
@@ -132,7 +134,7 @@ export const getRenderer = ({
     >
       <Parent>
         {column.format ? (
-          column.format(value)
+          column.format.format(value as any)
```
conglei, line 137: @kristw  this one I'm not so sure. I got type error if I don't add `.format`

packages/superset-ui-plugin-chart-table/src/Table.tsx
```diff
@@ -71,6 +72,44 @@ function getText(value: unknown) {
...
 }
 
+type columnWidthMetaDataType = {
+  [key: string]: {
+    maxWidth: number;
+    width: number;
+  };
+};
+
+const columnWidthSelector = createSelector<
```
kristw: Do you need to define all generics?

Usually I do this

```ts
createSelector(
  (a: AType) => a.b,
  a => a.c,
  (b, c) => ...
);
```
Only define the type of the first argument and it will infer correctly the rest.

kristw: **Important**

You have to define this selector as class function instead of outside. Otherwise this won't work if you have multiple instance of `TableVis`. They will share the same cache and keep invalidating.

Each table should have `this.columnWidthSelector` instead of single `columnWidthSelector`.

conglei:
@kristw just fixed the issues :)

packages/superset-ui-plugin-chart-table/src/processMetrics.ts
```diff
@@ -20,3 +23,10 @@ export default function processMetrics({
     .concat(processedPercentMetrics)
     .filter(m => typeof records[0][m] === 'number');
 }
+
+export default createSelector(
```
kristw: You better export a factory `() => createSelector(...)` so it can be used multiple times.

packages/superset-ui-plugin-chart-table/src/renderer.tsx
kristw: That's fine. There was some typing edge case that typescript fail to infer.

packages/superset-ui-plugin-chart-table/src/Table.tsx
```diff
@@ -71,7 +72,45 @@ function getText(value: unknown) {
...
+type columnWidthMetaDataType = {
+  [key: string]: {
+    maxWidth: number;
+    width: number;
+  };
+};
+
 class TableVis extends React.PureComponent<InternalTableProps, TableState> {
+  columnWidthSelector = createSelector<
+    ParentRow[],
```
kristw: Are the generic necessary? Is it possible to use the same way with other selectors that only the first argument type is defined?

kristw (approved):

kristw:
A few more minor points, but none blocking. Please fix and merge.

PR #247, opened 2019-11-05
fix: pass all props to transformProps in LineMulti chart

nytai:
💔 Breaking Changes

🏆 Enhancements

📜 Documentation

🐛 Bug Fix
This pr fixes an issue where LineMulti chart doesn't render with the current superset/master code. 

@mistercrunch 
🏠 Internal

### DIFF
--- a/packages/superset-ui-legacy-preset-chart-nvd3/src/LineMulti/LineMulti.jsx
+++ b/packages/superset-ui-legacy-preset-chart-nvd3/src/LineMulti/LineMulti.jsx
@@ -31,13 +31,12 @@ const propTypes = {
   annotationData: PropTypes.object,
   datasource: PropTypes.object,
   formData: PropTypes.object,
-  onAddFilter: PropTypes.func,
-  onError: PropTypes.func,
+  hooks: PropTypes.shape({
+    onAddFilter: PropTypes.func,
+    onError: PropTypes.func,
+  }),
 };
-const defaultProps = {
-  onAddFilter() {},
-  onError() {},
-};
+const defaultProps = {};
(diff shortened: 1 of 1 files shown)


packages/superset-ui-legacy-preset-chart-nvd3/src/LineMulti/LineMulti.jsx
```diff
@@ -31,8 +31,10 @@ const propTypes = {
...
   formData: PropTypes.object,
-  onAddFilter: PropTypes.func,
-  onError: PropTypes.func,
+  hooks: PropTypes.shape({
+    onAddFilter: PropTypes.func,
+    onError: PropTypes.func,
+  }),
 };
 const defaultProps = {
   onAddFilter() {},
```
kristw: You don't need this.now.

nytai: oh, didn't see that there. Wonder why is was failing on master if it has that fallback. 🤷‍♂

kristw: There is an ongoing issue with build tools, probably not related to this.

nytai:
Looks like the build is complaining with a babel error: `Cannot set property inList of [object Object] which has only a getter`

I was getting the same in dev when trying to build.

kristw:
Please rebase

kristw (approved):

mistercrunch:
🔨 🐛 🔥

PR #251, opened 2019-11-08
0.11.5

nytai:
💔 Breaking Changes

🏆 Enhancements

📜 Documentation
releasing: https://github.com/apache-superset/superset-ui-plugins/pull/247 
🐛 Bug Fix

🏠 Internal


kristw:
This should be handle as part of release script. No need to PR.

PR #253, opened 2019-11-12
fix(table): fix sorting, column width calculation, and text wrapping

conglei:
💔 Breaking Changes

🏆 Enhancements

📜 Documentation

🐛 Bug Fix
This PR is to

1) better calculation of the column width (with formatted string);
image

2) lexicographically sorting strings;
image

3) fixing word truncating issues;
image

🏠 Internal

cc: @etr2460 @graceguo-supercat

### DIFF
--- a/packages/superset-ui-plugin-chart-table/src/Table.tsx
+++ b/packages/superset-ui-plugin-chart-table/src/Table.tsx
@@ -3,9 +3,16 @@ import DataTable from '@airbnb/lunar/lib/components/DataTable';
 import Text from '@airbnb/lunar/lib/components/Text';
 import Input from '@airbnb/lunar/lib/components/Input';
 import withStyles, { WithStylesProps } from '@airbnb/lunar/lib/composers/withStyles';
-import { Renderers, ParentRow, ColumnMetadata } from '@airbnb/lunar/lib/components/DataTable/types';
+import {
+  Renderers,
+  ParentRow,
+  ColumnMetadata,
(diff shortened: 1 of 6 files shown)


packages/superset-ui-plugin-chart-table/src/Table.tsx
```diff
@@ -60,16 +67,20 @@ function getCellHash(cell: Cell) {
   return `${cell.key}#${cell.value}`;
 }
 
-function getText(value: unknown) {
+function getText(value: unknown, format: TimeFormatter | NumberFormatter | undefined) {
+  let formattedValue = value;
+  if (format) {
+    formattedValue = format.format(value as any);
```
kristw: Should we return here? If there is a `format` function, that should not produce html output.

packages/superset-ui-plugin-chart-table/src/Table.tsx
```diff
@@ -60,16 +67,20 @@ function getCellHash(cell: Cell) {
...
+  }
   if (typeof value === 'string') {
     const span = document.createElement('span');
```
kristw: Another new trick I learn is using regex to remove all html tags. 
I think it should be faster than parsing html via `span` tag, although haven't really tested it.

```ts
const htmlTagRegex = /(<([^>]+)>)/gi;
const text = value.replace(htmlTagRegex, '');
```

kristw (approved):

PR #258, opened 2019-11-16
feat(plugin-chart-word-cloud): convert word cloud to use encodable

kristw:
🏆 Enhancements

* Define WordCloud's API using `encodable`.

**Basic**
image

**encodes color by word length**
image

**encodes font family by first letter**
image


packages/superset-ui-plugin-chart-word-cloud/src/chart/WordCloud.tsx
```diff
@@ -0,0 +1,128 @@
+import React from 'react';
+import cloudLayout, { Word } from 'd3-cloud';
+import { PlainObject } from 'encodable/lib/types/Data';
+import { WordCloudEncoding, wordCloudEncoderFactory } from './Encoder';
+
+export const ROTATION = {
+  flat: () => 0,
+  /* eslint-disable-next-line no-magic-numbers */
+  random: () => Math.floor(Math.random() * 6 - 3) * 30,
```
espressoroaster, line 10: Maybe leave a comment to explain that this calculates a random rotation between -90 and 90 degrees. I'm assuming the unit is degrees since square is `* 90`, but it would be helpful to read it explicitly when skimming the code.

espressoroaster (commented):
I've only taken a look at Encoder.ts and Wordcloud.tsx, but those two LGTM. Have a non-blocking comment about adding a comment for the rotation function.

PR #276, opened 2019-12-02
Gh 8669 tooltip overflow fix

rusackas:
🏆 Enhancements
🐛 Bug Fix

This addresses Superset issue 8669:
https://github.com/apache/incubator-superset/issues/8669

It also makes the headers bold, which hopefully looks better to folks.

Before:
image

After:
image


etr2460:
I think you need to fix the commit messages first. I'd recommend squashing them all into one and following the conventional commit guidelines

PR #278, opened 2019-12-02
fix: GH issue 8669 NVD3 tooltip overflow

rusackas:
🐛 Bug Fix

This addresses Superset issue 8669:
https://github.com/apache/incubator-superset/issues/8669

It also makes the headers bold, which hopefully looks better to folks.

Before:
image

After:
image


packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.css
```diff
@@ -121,3 +121,13 @@
...
+
+/* START tooltip styles - these prevent long column names from overflowing the tooltip */
+.nvtooltip tbody tr:first-child td{
+  white-space: nowrap;
+  font-weight: bold;
+}
+.nvtooltip tbody tr:not(:first-child) td:nth-child(2){
+  word-break: break-word;
+}
+/* END tooltip styles */
```
etr2460: nit: add a new line

etr2460 (approved):
lgtm with one nit

packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.css
rusackas: Committed. I wish people didn't object to using Husky to add pre-commit hooks running Prettier

williaster: who objects to husky pre-commit hooks? I support that 👍

williaster: on second thought, I'm not so sure. we've had these in other repos and  the major downside of pre-commit hooks is that they can take longer and longer to run, especially as the repo grows. 

I think you can setup your `ide` to auto-fix eslint (which calls prettier) changes on save.

PR #279, opened 2019-12-02
fix: apply margins in nvd3

### LINKED ISSUE #8480: Data labels overlap with legend once again
As described here #3779 and fixed last year by #6145, labels of tall bars overlap with the legend. However, the file modified by #6145 exists only until branch 0.27. Wasn't able to figure out which file to address to fix this in master, sorry.

### Expected results

Bar labels and legend not overlapping.

### Actual results

Bar labels overlapping with legend.

#### Screenshots

67417267-a7439800-f5c8-11e9-853f-48a6df86ad30

#### How to reproduce the bug

1. Create some fake numeric data, can be all of the same values, just make them large enough (like an array of "1000").
2. Create bar chart from this data, enable legend, enable bar values.
3. Either already in the editor or after exporting to dashboard you'll see labels overlapping the legend.

Screenshot attached for reference anyway.

### Environment

- superset version: 0.34.1
- python version: 3.7.3
- node.js version: 10.16.0
- npm version: 6.10.2

### Checklist

- [✓] I have checked the superset logs for python stacktraces and included it here as text if there are any.
- [✓] I have reproduced the issue with at least the latest released version of superset.
- [✓] I have checked the issue tracker for the same issue and I haven't found one similar.
^ Actually I have but it is 1 year old and the bug reappeared despite it being supposedly fixed in a previous version.

betodealmeida:
🐛 Bug Fix

The NVD3 visualization does a bunch of operations on the margin object, depending on the visualization parameters. For example, if we're showing the labels on bars we need to increase the top margin by 24px so that it doesn't overlap with the legend:

```javascript
      if (showBarValue) {
        // Add more margin to avoid label colliding with legend.
        margins.top += 24;
      }
```

The problem is that the `margins` object is never assigned to the chart. This PR fixes it.

Before:

Screen Shot 2019-12-02 at 3 30 32 PM

After:

Screen Shot 2019-12-02 at 3 30 01 PM

This PR fixes https://github.com/apache/incubator-superset/issues/8480.

### DIFF
--- a/packages/superset-ui-legacy-plugin-chart-parallel-coordinates/src/vendor/parcoords/d3.parcoords.js
+++ b/packages/superset-ui-legacy-plugin-chart-parallel-coordinates/src/vendor/parcoords/d3.parcoords.js
@@ -1759,10 +1759,9 @@ module.exports = function(config) {
       var svg = pc.selection.select('svg').select('g#arcs'),
         id = arc.dims.i,
         points = [arc.p2, arc.p3],
-        line = svg.selectAll('line#arc-' + id).data([
-          { p1: arc.p1, p2: arc.p2 },
-          { p1: arc.p1, p2: arc.p3 },
-        ]),
+        line = svg
+          .selectAll('line#arc-' + id)
+          .data([{ p1: arc.p1, p2: arc.p2 }, { p1: arc.p1, p2: arc.p3 }]),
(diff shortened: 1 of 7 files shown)


etr2460:
Once you rebase, the build should be fixed

etr2460:
also, it looks like a bunch of extra formatting changes were applied here. could you remove those? (maybe you're on an old version of prettier?)

PR #171, opened 2019-08-09
Fix D3 format for some viz types

spotx-ajacobi:
Some viz types don't honor the predefined D3 formats users enter for metrics. This change fixes this for a handful of viz types.

PR #280, opened 2019-12-02
fix: Making viz components respect D3 Format from metric 

rusackas:
🐛 Bug Fix

Addresses this bug:
https://github.com/apache/incubator-superset/issues/7650

Riffs on: 
https://github.com/apache-superset/superset-ui-plugins/pull/171/files

Screenshots with a goofy number format to prove the point:

Bar (before):
image

Bar (after):
image

Line (before):
image

Line (after):
image

Pie (before):
image

Pie (after):
image

Dual line (before):
image

Dual line (after):
image

Area (before):
image

Area (after):
image

Treemap (before):
image

Treemap (after):
image

# Reviewers:
@kristw @etr2460 @mistercrunch


packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/transformProps.js
```diff
@@ -81,8 +84,16 @@ export default function transformProps(chartProps) {
   } else if (percentChange < 0) {
     className = 'negative';
   }
+  let metricFormat = yAxisFormat;
+  if (!yAxisFormat && chartProps.datasource && chartProps.datasource.metrics) {
```
etr2460, line 88: i think i'd prefer `!metricFormat` here

packages/superset-ui-legacy-preset-chart-nvd3/src/transformProps.js
```diff
@@ -79,6 +78,26 @@ export default function transformProps(chartProps) {
       }))
     : rawData;
 
+  const grabD3Format = targetMetric => {
```
etr2460: let's define this outside of the block

packages/superset-ui-legacy-preset-chart-nvd3/src/transformProps.js
```diff
@@ -79,6 +78,26 @@ export default function transformProps(chartProps) {
...
+    return foundFormatter;
+  };
+
+  if (chartProps.formData.vizType === 'pie') {
+    numberFormat = grabD3Format(chartProps.formData.metric) || numberFormat;
+  } else if (chartProps.formData.vizType === 'dual_line') {
+    yAxisFormat = grabD3Format(chartProps.formData.metric) || yAxisFormat;
+    yAxis2Format = grabD3Format(chartProps.formData.metric2) || yAxis2Format;
+  } else if (['line', 'dist_bar', 'bar', 'area'].indexOf(chartProps.formData.vizType) > -1) {
+    yAxisFormat = grabD3Format(chartProps.formData.metrics[0]) || yAxisFormat;
```
etr2460: why do we overwrite the passed in format here but we don't in the other charts?

etr2460 (changes requested):
some questions

packages/superset-ui-legacy-preset-chart-nvd3/src/transformProps.js
rusackas: I just spent a bit more time looking at this one, honestly. I think overwriting is more straightforward - I'll make the others conform, which also addresses your earlier comment about the (unnecessary, really) `metricFormat` variable.

rusackas: Now that I'm looking at this with fresh eyes, I was a little heavy handed in this approach, and I'm making a tweak. 

With this logic:
`yAxisFormat = grabD3Format(chartProps, chartProps.formData.metrics[0]) || yAxisFormat;`
... the `D3 Format` from the metric actually overrides the form data for the viz. That's not cool. I'll recommit as:
`yAxisFormat = yAxisFormat || grabD3Format(chartProps, chartProps.formData.metrics[0]);`
... then the formData (if present) will act as an override of the Metric's D3 Format setting. That setting can be cleared in the UI, allowing the Metric's format to show itself. If Neither the form nor the Metric specify a format, NVD3 takes its best guess.

packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/transformProps.js
rusackas: `metricFormat` wasn't really necessary here... I made this one overwrite the passed in `yAxisFormat` like the others do. I think that's more straightforward.

packages/superset-ui-legacy-preset-chart-nvd3/src/transformProps.js
rusackas: done.

rusackas:
> some questions

@etr2460 I think everything here was addressed and/or improved upon (thanks for looking at it!) - let me know if there's anything else that looks like it needs attention.

rusackas:
@etr2460 / @mistercrunch wondering if you wouldn't mind a (re)review on this.

mistercrunch (approved):

PR #286, opened 2019-12-04
fix: TTestTable viz had className error, and broken CSS styling

rusackas:
🐛 Bug Fix

There was a method returning a boolean used as a className, causing two issues:
1) The console was barking about using the boolean at all
2) The CSS for .true and .false classes was not being applied for lack of a string className.

This easy peasy fix addresses both issues.

Before:
image

After:
image


etr2460 (approved):
makes sense, lgtm!

PR #287, opened 2019-12-04
fix: JS exception: Cannot read property 'dispatch' of undefined

graceguo-supercat:
🐛 Bug Fix

Sometimes chart took longer time to render, and annotation layer is rendered before chart is ready. In this case, when the annotation layer tried to attach event handler for brush events, it will cause JS exception because chart.focus object is not created:
Screen Shot 2019-12-04 at 2 58 57 PM

This error will make chart section blank, and cannot apply annotation at all.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 *.log
 *.map
 *.min.js
+.*.swp
 
 babel.config.js
 build/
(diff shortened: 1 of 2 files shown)


etr2460 (approved):
lgtm, thanks!

PR #291, opened 2019-12-10
fix: nvd3 tooltip bolding/non-linebreak on ONLY header row

rusackas:
🐛 Bug Fix

In NVD3 charts, legends were getting for their first row, thinking it was a header. This header row is not always a header! Now, the template gets a class for the header row, which we can leverage in CSS, so headerless charts are unaffected.

Before (no header row, no problem):
image

After (with header row, still no problem):
image

Before (no header row, PROBLEM!):
Pasted_Image_12_9_19__5_19_PM

After (no header row, NO problem):
image

Review, please :) @etr2460


packages/superset-ui-legacy-preset-chart-nvd3/src/utils.js
```diff
@@ -143,7 +143,7 @@ export function generateAreaChartTooltipContent(d, timeFormatter, valueFormatter
     "<table><thead><tr><td colspan='4'>" +
     `<strong class='x-value'>${timeFormatter(d.value)}</strong>` +
     '</td></tr></thead><tbody>' +
-    '<tr><td></td><td>Category</td><td>Value</td><td>% to total</td></tr>';
+    '<tr class="legend-header"><td></td><td>Category</td><td>Value</td><td>% to total</td></tr>';
```
etr2460: this should probably be `tooltip-header` or maybe `tooltip-column-names`? I don't think it's technically a legend

rusackas: Agreed! `tooltip-header` it is!

etr2460 (approved):
lgtm! we're working on fixing the travis issues now, will merge once they're resolved

rusackas:
@etr2460 Let me know if you see something I can address/do to make our friend Travis happier :)

rusackas:
@etr2460 any way to retry Travis on a PR like this, or should I find some random line of code to commit and see if it runs?

mistercrunch:
Try rebasing (?)

rusackas:
Travis failed again... not quite sure why, as I have not been successful in getting builds to fail locally.

kristw (approved):

kristw:
Please rebase

PR #312, opened 2020-01-16
Fix pivot table setup verbose name

ma0qian0:
🐛 Bug Fix


ma0qian0:
Should check and replace all verbose names

kristw:
Please rebase

PR #319, opened 2020-01-30
build: migrate to nimbus and fix build

kristw:
🏠 Internal

* Migrate build tools from `beemo` to `nimbus`
* Fix issues
  * Move `buildAssets.js` under `scripts`
  * Replace `eslint` rule `sort-keys` with `jsx-sort-default-props` and `sort-prop-types`
  * Remove unnecessary `eslint-disable`
  * Replace `if (x.length)` with `if (x.length > 0)`
  * Switch `for` loop to use `.forEach` when possible
  * Change `module.exports` to `export default` (for parallel coordinates)
  * Fix other lint errors

Tested by manually going through all examples in storybook and verify them against `master`. 

@etr2460 @rusackas


kristw:
Ready for review

package.json
```diff
@@ -4,25 +4,27 @@
...
-    "prepare-release": "git checkout master && git pull --rebase origin master && lerna bootstrap && yarn run test",
-    "prerelease": "yarn run build",
-    "pretest": "yarn run lint",
-    "prettier": "beemo prettier \"./packages/*/{src,test,storybook}/**/*.{js,jsx,ts,tsx,json,md}\"",
-    "release": "yarn run prepare-release && lerna publish && yarn run postrelease",
-    "test": "yarn run type && yarn run jest",
-    "test:watch": "yarn run lint:fix && beemo create-config jest --react && jest --watch"
+    "format": "yarn prettier --write",
+    "jest": "NODE_ENV=test nimbus jest --coverage --verbose",
+    "lint": "nimbus eslint",
```
rusackas: If this is the task run by CI, should we validate pretty-ness here, e.g.:
```suggestion
    "lint": "nimbus eslint && nimbus prettier --check",
```

kristw: I think the commit hook with `husky` should auto format, but doesn't hurt to double check I guess.

package.json
```diff
@@ -4,25 +4,27 @@
...
-    "test": "yarn run type && yarn run jest",
-    "test:watch": "yarn run lint:fix && beemo create-config jest --react && jest --watch"
+    "format": "yarn prettier --write",
+    "jest": "NODE_ENV=test nimbus jest --coverage --verbose",
+    "lint": "nimbus eslint && nimbus prettier --check",
+    "lint:fix": "yarn lint --fix",
+    "prettier": "nimbus prettier",
+    "test": "yarn type:dts && yarn jest",
+    "test:watch": "yarn lint:fix && yarn jest --watch",
+    "type:dts": "nimbus typescript --build --reference-workspaces",
```
rusackas: nit/question - I noticed the `type` script has been removed - should `type:dts` be renamed to `type` just to simplify?

packages/superset-ui-legacy-plugin-chart-heatmap/src/Heatmap.css
```diff
@@ -43,13 +43,14 @@
...
+.superset-legacy-chart-heatmap img {
+  image-rendering: optimizeSpeed; /* Older versions of FF          */
+  image-rendering: -moz-crisp-edges; /* FF 6.0+                       */
   image-rendering: -webkit-optimize-contrast; /* Safari                        */
-  image-rendering: -o-crisp-edges;            /* OS X & Windows Opera (12.02+) */
-  image-rendering: pixelated;                 /* Awesome future-browsers       */
-  -ms-interpolation-mode: nearest-neighbor;   /* IE                            */
+  image-rendering: -o-crisp-edges; /* OS X & Windows Opera (12.02+) */
+  image-rendering: pixelated; /* Awesome future-browsers       */
+  -ms-interpolation-mode: nearest-neighbor; /* IE                            */
```
rusackas, line 53: Surprised Prettier isn't smart enough to convert these to `//` comments!

packages/superset-ui-legacy-plugin-chart-map-box/src/ScatterPlotGlowOverlay.jsx
```diff
@@ -151,7 +150,7 @@ class ScatterPlotGlowOverlay extends React.PureComponent {
           ctx.beginPath();
           if (location.get('properties').get('cluster')) {
             let clusterLabel = clusterLabelMap[i];
-            const scaledRadius = roundDecimal(Math.pow(clusterLabel / maxLabel, 0.5) * radius, 1);
+            const scaledRadius = roundDecimal((clusterLabel / maxLabel) ** 0.5 * radius, 1);
```
rusackas: This is fine, but just as food for thought, this instance of ** is a slightly less friendly than the first use on line 63. In this use case, the reader has to consiously think about order of operations, whereas Math.pow(...) makes that pretty explicit.

packages/superset-ui-legacy-plugin-chart-map-box/src/utils/geo.js
```diff
@@ -8,7 +7,7 @@ export function kmToPixels(kilometers, latitude, zoomLevel) {
   // Algorithm from: http://wiki.openstreetmap.org/wiki/Zoom_levels
   const latitudeRad = latitude * (Math.PI / 180);
   // Seems like the zoomLevel is off by one
-  const kmPerPixel = (EARTH_CIRCUMFERENCE_KM * Math.cos(latitudeRad)) / Math.pow(2, zoomLevel + 9);
+  const kmPerPixel = (EARTH_CIRCUMFERENCE_KM * Math.cos(latitudeRad)) / 2 ** (zoomLevel + 9);
```
rusackas:
```suggestion
  const kmPerPixel = (EARTH_CIRCUMFERENCE_KM * Math.cos(latitudeRad)) / (2 ** (zoomLevel + 9));
```
Again totally optional, but I wonder if parens make this easier to read.

packages/superset-ui-legacy-plugin-chart-markup/src/Markup.css
```diff
@@ -26,7 +26,13 @@
 .separator hr {
   border: 0;
   height: 1px;
-  background-image: linear-gradient(to right, rgba(0, 0, 0, 1), rgba(0, 0, 0, 1), rgba(0, 0, 0, 1), rgba(0, 0, 0, 0));
+  background-image: linear-gradient(
+    to right,
```
rusackas, line 30: Could simplify this (if you indeed think it's simpler) to something like:
```
background-image: linear-gradient(
    to right,
    rgba(0, 0, 0, 1) 66%,
    rgba(0, 0, 0, 0)
  );
```
example in action: https://codepen.io/rusackas-1470804400/pen/oNXvdYW

packages/superset-ui-legacy-plugin-chart-parallel-coordinates/src/vendor/parcoords/d3.parcoords.css
```diff
@@ -41,22 +43,26 @@
...
-	-khtml-user-select: none;
-	-moz-user-select: none;
-	-ms-user-select: none;
-	user-select: none;
-    background-color: white;
+  -webkit-touch-callout: none;
+  -webkit-user-select: none;
+  -khtml-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
```
rusackas, line 50: Just realizing we're not using autoprefixer... I'll try to make a PR around that soon :D

packages/superset-ui-legacy-plugin-chart-heatmap/src/Heatmap.css
williaster: I don't think that works in `css`

packages/superset-ui-legacy-plugin-chart-map-box/src/utils/geo.js
williaster: I think prettier removes them so would have to `prettier-ignore`

packages/superset-ui-legacy-plugin-chart-sunburst/src/Sunburst.js
```diff
@@ -312,8 +311,10 @@ function Sunburst(element, props) {
         if (!isLeafNode) {
           // Not yet at the end of the sequence; move down the tree.
           let foundChild = false;
-          for (let k = 0; k < children.length; k++) {
-            currChild = children[k];
+
+          // eslint-disable-next-line unicorn/no-for-loop
+          for (let i = 0; i < children.length; i++) {
```
rusackas: If you did want to get rid of this for loop, you could probably also get rid of the `foundChild` variable with something like this:
```
childNode = children.filter(child => child.name === nodeName && child.level === level)[0]; // returns the first value or falsey undefined

if(!childNode){
  childNode = {
    name: nodeName,
    children: [],
    level,
  }
  children.push(childNode);
}
```

packages/superset-ui-legacy-plugin-chart-sunburst/src/utils/wrapSvgText.js
```diff
@@ -29,8 +29,8 @@ export default function wrapSvgText(text, width, adjustedY) {
       .attr('dy', `${dy}em`);
 
     let didWrap = false;
-    for (let i = 0; i < words.length; i++) {
-      word = words[i];
+    words.forEach(element => {
```
rusackas: this is fine, but you could potentially not declare `word` above, and just `forEach(word => ...)` directly

packages/superset-ui-legacy-preset-chart-nvd3/src/utils/isTruthy.js
```diff
@@ -1,8 +1,9 @@
 export default function isTruthy(obj) {
   if (typeof obj === 'boolean') {
     return obj;
-  } else if (typeof obj === 'string') {
-    return ['yes', 'y', 'true', 't', '1'].indexOf(obj.toLowerCase()) >= 0;
+  }
+  if (typeof obj === 'string') {
+    return ['yes', 'y', 'true', 't', '1'].includes(obj.toLowerCase());
```
rusackas, line 6: Not sure if the linter complained about this, but I kind of like `else if` stylistically, plus it's one less computation in some cases, but I won't lose any sleep over it :D

packages/superset-ui-legacy-preset-chart-nvd3/test/utils.test.js
```diff
@@ -72,12 +72,12 @@ const DATA_WITH_DISABLED_SERIES = [
     key: ['East Asia & Pacific'],
     values: [
       {
-        x: -315619200000.0,
```
rusackas: Not sure if there was a valid reason these are `.0` like to cast as float and make sure it comes through a formatter correctly.

packages/superset-ui-plugin-chart-icicle-event/src/IcicleEventChart.tsx
```diff
@@ -129,7 +129,7 @@ export default class IcicleEventChart extends Component<Props> {
       .attr('width', d => rectWidth(isVertical, boxMargin, d))
       .attr('height', d => rectHeight(isVertical, boxMargin, d))
       .attr('rx', rounding)
-      .attr('fill', d => color(d.data.name || ''))
+      .attr('fill', d => color(d.data.name ?? ''))
```
rusackas, line 132: 😲 nice

packages/superset-ui-plugin-chart-icicle-event/src/utils/IcicleEventTreeHelpers.ts
```diff
@@ -32,14 +32,14 @@ export function findDepth(node: IcicleEventNode, depth: number = 0): number {
...
-  return maxDepth || depth;
+  return maxDepth ?? depth;
 }
 
 export function hierarchySort(
   a: HierarchyNode<IcicleEventNode>,
   b: HierarchyNode<IcicleEventNode>,
 ): number {
-  if (a && a.value && b && b.value) {
+  if (a?.value && b && b.value) {
```
rusackas: why not `b?.value` as well?

packages/superset-ui-plugin-chart-table/src/components/HTMLRenderer.tsx
```diff
@@ -4,14 +4,15 @@ import dompurify from 'dompurify';
 const isHTML = RegExp.prototype.test.bind(/(<([^>]+)>)/i);
 
 export default function HTMLRenderer({ value }: { value: string }) {
-  if (isHTML(value)) {
-    const html = useMemo(() => ({ __html: dompurify.sanitize(value) }), [value]);
+  const html = useMemo(() => ({ __html: dompurify.sanitize(value) }), [value]);
```
rusackas: Just for my own edification, was this moved out for linting reasons, or some other practical reason? Seems like it's unnecessary to do the `useMemo` here if `value` turns out not to be HTML. ¯\\\_(ツ)_/¯

rusackas (approved):
LGTM, but I've peppered the thread with little comments and questions about various nits in case you think any are worth addressing. Love all the cleanup work you've done in here, and it's nice to see newer ES features coming in to play!

packages/superset-ui-legacy-plugin-chart-heatmap/src/Heatmap.css
rusackas: Hah, you're right... everything is (LE/SA)SS in my head by default :P

packages/superset-ui-legacy-plugin-chart-markup/src/Markup.css
kristw: This was done by prettier. May have to change the config.

rusackas: Not important, no worries.

packages/superset-ui-plugin-chart-icicle-event/src/utils/IcicleEventTreeHelpers.ts
kristw: These are changed by `eslint` auto-fix. Perhaps it is not smart enough. I'll take a look.

package.json
kristw: good idea

packages/superset-ui-legacy-plugin-chart-map-box/src/ScatterPlotGlowOverlay.jsx
kristw: bringing back `Math.pow`

packages/superset-ui-legacy-preset-chart-nvd3/src/utils/isTruthy.js
kristw: The linter complains about it

packages/superset-ui-legacy-preset-chart-nvd3/test/utils.test.js
kristw: Added `eslint-disable` for the entire file and bring back `.0`

packages/superset-ui-legacy-plugin-chart-sunburst/src/Sunburst.js
kristw: Great idea. I made small change replacing `.filter` with `.find`

PR #327, opened 2020-02-06
feat: support no data within BigNumber viz

etr2460:
🏆 Enhancements

Support an empty array being passed in as data. This is precursor work to making "No data" not an error within Superset.

Screen Shot 2020-02-06 at 2 46 45 PM

to: @kristw


kristw:
I think code-wise looks good. 

Would you add checks like this to all charts? Would be nice to have some design consistency. 
Wonder if design would have any suggestion.

etr2460:
The plan is to make all charts support this, and keep the design consistent with however the chart looks today. For example, all NVD3 charts already have a no data available treatment, so i'll keep that around. For other types of charts (like geovis) i'd probably keep the map visible and overlay the no data message

kristw (approved):

PR #330, opened 2020-02-07
feat: customize no data message in nvd3 charts

etr2460:
🏆 Enhancements

Update the no data message for nvd3 charts. Unfortunately, nvd3 only supported a single line of text for the no data message, so i overrode the noData function to support a custom, multiline message. Not sure if this is the best way to do this, definitely open to other suggestions!

Screen Shot 2020-02-07 at 11 46 09 AM

to: @kristw @rusackas

### DIFF
--- a/packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
+++ b/packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
@@ -67,6 +67,38 @@ import {
 } from './PropTypes';
 import './NVD3Vis.css';
 
+// Override the noData render function to make a prettier UX
+// Code adapted from https://github.com/novus/nvd3/blob/master/src/utils.js#L653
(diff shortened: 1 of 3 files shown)


packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.js
```diff
@@ -67,6 +67,38 @@ import {
 } from './PropTypes';
 import './NVD3Vis.css';
 
+// Override the noData render function to make a prettier UX
+// Code adapted from https://github.com/novus/nvd3/blob/master/src/utils.js#L653
+nv.utils.noData = function(chart, container) {
+  const opt = chart.options();
+  const margin = opt.margin();
+  const data = [
```
kristw: Can move this constant outside.

kristw (approved):
One minor change.

kristw (approved):

PR #372, opened 2020-02-25
fix: add container css for interval type annotation

graceguo-supercat:
🐛 Bug Fix

Airbnb user reported an issue that Interval type annotation overlapped with chart labels.
It seems the old UI look like this:
Screen Shot 2019-10-17 at 3 29 03 PM (1)
This issue was reported pretty long time ago so I can't trace back since when we lost this container css. So I just added container class and some new css.

Before:
Screen_Shot_2020-02-25_at_3_08_52_PM

After:
Screen Shot 2020-02-25 at 4 20 15 PM

@etr2460 @kristw @ktmud

### DIFF
--- a/packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.css
+++ b/packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.css
@@ -166,9 +166,9 @@
 
 .d3-tip.nv-event-annotation-layer-table::after {
   content: '\25BC';
-  font-size: 22px;
+  font-size: 14px;
   color: #484848;
   position: absolute;
-  bottom: -23px;
-  left: 84px;
+  bottom: -14px;
+  left: 94px;
 }
--- a/packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.css
+++ b/packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.css
@@ -153,3 +153,22 @@
   font-size: 14px;
   font-weight: normal;
 }
+
+.d3-tip.nv-event-annotation-layer-table {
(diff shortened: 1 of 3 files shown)


packages/superset-ui-legacy-preset-chart-nvd3/src/utils.js
```diff
@@ -295,7 +295,7 @@ export function wrapTooltip(chart, maxWidth) {
 
 export function tipFactory(layer) {
   return d3tip()
-    .attr('class', 'd3-tip')
+    .attr('class', `d3-tip ${layer.annotationTipClass}`)
```
ktmud: Better use `{layer.annotationTipClass || ''}` so we don't add an "undefined" class to the tip.

graceguo-supercat: fixed.

packages/superset-ui-legacy-preset-chart-nvd3/src/NVD3Vis.css
```diff
@@ -153,3 +153,22 @@
...
+}
+
+.d3-tip.nv-event-annotation-layer-table::after {
+  content: '\25BC';
+  font-size: 22px;
+  color: #484848;
+  position: absolute;
+  bottom: -23px;
+  left: 84px;
+}
```
ktmud: Here seems to miss a new line. The arrow also looks larger than the old UI. Everything else looks good to me!

graceguo-supercat: fixed new line. also make down arrow smaller. thanks!

kristw (approved):

PR #385, opened 2020-02-29
perf: faster legacy table chart

ktmud:
This commit tries to optimize the performance of the legacy data table.

1. Still use the jquery.databables plugin.
2. Instead of `d3` and `reactify`, we create the DOM with React itself and initialize `jquery.DataTables` with a `useEffect` hook.
3. Removed dependency on `dompurify`. Now that HTML tags will be rendered as plain text (by React) instead of `innerHTML`.
4. Removed a couple of not-so-popular features. See notes below.
5. Added a script to build only specific plugin for faster development.
6. Restyled the pagination so it looks nicer in dashboards.

Plus some minor changes to fix linting and building.

Related: https://github.com/apache/incubator-superset/pull/9234

## Screenshots

Tested with a ~10,000 rows, 7 columns table in Superset, pagination enabled:

### Before 

legacy-table--old

Takes about 6 secs to load.

Snip20200303_3

### After

legacy-table--new

Takes only 3 secs to load!

Snip20200303_4

🏆 Enhancements


package.json
```diff
@@ -159,5 +159,8 @@
       "yarn prettier --write",
       "git add"
     ]
+  },
+  "resolutions": {
+    "**/@types/react": "^16.9.23"
```
ktmud, line 164: Had to specify `@types/react` version because `interweave-emoji` from `lunar` somehow [picked a fixed version](https://github.com/milesj/interweave/commit/0418313060f88b196ed59dfc0c5b55b6028205a5#r37505021) and it's throwing errors in the latest TypeScript.

packages/superset-ui-plugins-demo/package.json
```diff
@@ -39,7 +39,8 @@
     "@storybook/react": "^5.0.9",
     "@types/react": "^16.8.8",
     "@types/storybook__react": "4.0.2",
-    "bootstrap": "^4.3.1",
+    "bootstrap": "^3.4.1",
```
ktmud, line 42: Current Superset is still on Bootstrap 3.

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
```diff
@@ -0,0 +1,231 @@
...
+    includeSearch = false,
+    metrics: aggMetrics,
+    pageLength,
+    percentMetrics,
+    tableTimestampFormat,
+    // orderDesc,
+    // TODO: add back the broken dashboard filters feature
+    // filters = {},
+    // onAddFilter = NOOP,
+    // onRemoveFilter = NOOP,
```
ktmud, line 59: The table filters feature (click on table cells to add a filter) is still broken in production. Discussed with @graceguo-supercat , it's better to skip implementing this feature for now. Might add back later in another PR.

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
```diff
@@ -0,0 +1,231 @@
...
+    // tableFilter,
+    // timeseriesLimitMetric,
```
ktmud, line 61: `timeseriesLimitMetric` is used to sort table by timestamp when present. Didn't think this is very useful (there are many ways to sort the input data already), so didn't implement in this PR either.

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
```diff
@@ -0,0 +1,231 @@
...
+      isMetric,
+      className,
+      title: label,
+      data: key,
+      render: renderCell,
+    };
+  });
+
+  const viewportHeight = Math.min(height, window.innerHeight);
+  const pageLengthChoices = [10, 25, 40, 50, 75, 100, 150, 200];
```
ktmud, line 91: This is in consistent with [superset-frontend/src/explore/controls.js](https://github.com/apache/incubator-superset/blob/291306392443a5a0d0e2ee0cc4a95d37c56d4589/superset-frontend/src/explore/controls.jsx#L1336). Hard-code for now, constantize when refactoring controls to plugins.

packages/superset-ui-plugins-demo/.storybook/storybook.css
```diff
@@ -6,3 +6,7 @@ body,
   font-weight: 200;
   color: #484848;
 }
+
+#root > div {
+  padding: 8px;
+}
```
ktmud, line 11: Add a padding so it looks more close to an embedded Superset chart in dashboards.

packages/superset-ui-plugins-demo/storybook/stories/legacy-plugin-chart-table/Stories.tsx
```diff
@@ -1,9 +1,22 @@
 /* eslint-disable no-magic-numbers */
 import React from 'react';
 import { SuperChart } from '@superset-ui/chart';
+import { Props as SuperChartProps } from '@superset-ui/chart/lib/components/SuperChart';
 import data from './data';
+import birthNames from './birth_names.json';
```
ktmud, line 6: Large example data queried from Superset example dataset populated by `superset load_examples`.

packages/superset-ui-preset-chart-xy/src/encodeable/AxisAgent.ts
```diff
@@ -133,7 +133,7 @@ export default class AxisAgent<Def extends ChannelDef> {
 
     const { labelOverlap, labelPadding, orient } = this.config;
 
-    const maxWidth = Math.max(...tickLabelDimensions.map(d => d.width), 0);
+    const maxWidth = Math.max(...tickLabelDimensions.map((d: Dimension) => d.width), 0);
```
ktmud, line 136: Needed to pass Typescript linting.

packages/superset-ui-legacy-plugin-chart-table/src/Table.css
```diff
@@ -17,17 +17,25 @@
  * under the License.
  */
 .superset-legacy-chart-table {
-  margin: 0px !important;
+  margin: 0px auto;
```
ktmud, line 20: Not sure why there is an `!important` here but it doesn't seem warranted. Tested in Superset and it worked fine.

ktmud (commented):
Some notes.

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
```diff
@@ -177,22 +151,28 @@ export default function ReactDataTable(props: DataTableProps) {
...
 
+  /**
+   * Formatted text for cell value
+   */
+  function cellText(key: string, format: string | undefined, val: unknown) {
+    if (key === '__timestamp') {
+      return formatTimestamp(val);
+    } else if (typeof val === 'string') {
+      // It's fine to return raw value react will handles HTML escaping
+      return val;
```
kristw: Could we also test if the cells can render html links correctly?

kristw (commented):
> Removed dependency on `dompurify`. Now that HTML tags will be rendered as plain text (by React) instead of innerHTML.

There are columns values that are html strings, such as html links. You would need `dangerouslySetInnerHTML` and `dompurify` or use `interweave`.

kristw:
image

Could we add bottom padding for the pagination?
Also, could you share a screenshot of how does it look in Superset? I think Superset override some css and the table may look slightly different.

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
```diff
@@ -0,0 +1,238 @@
...
+    } else if (isMetric(key)) {
+      // default format '' will return human readable numbers (e.g. 50M, 33k)
+      return getNumberFormatter(format || '')(val as number);
```
kristw: Can do
```ts
import { formatNumber } from '@superset-ui/number-format';
```

```ts
return formatNumber(format || '', val as number);
```

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
ktmud: Updated with [js-xss](https://github.com/leizongmin/js-xss) library.

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
```diff
@@ -35,19 +36,9 @@ if (!dt.$) {
   dt(window, $);
 }
 
+const RE_HTML_TAG = /<.*>/; // a dead simple regexp to find html tag
```
kristw: Let's rename this to `htmlTagRegex` for clarity

kristw: Can make small adjustment to the regex.

`/(<([^>]+)>)/gi`

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
```diff
@@ -80,17 +71,14 @@ export default function ReactDataTable(props: DataTableProps) {
 
   const maxes: { [key: string]: number } = {};
   const mins: { [key: string]: number } = {};
-  const isMetric = (key: string) => metrics.includes(key);
+  // a fater way of determine whether a key is a metric
```
kristw: typo "faster"

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
```diff
@@ -162,11 +162,10 @@ export default function ReactDataTable(props: DataTableProps) {
     if (key === '__timestamp') {
       return formatTimestamp(val);
     } else if (typeof val === 'string') {
-      // It's fine to return raw value react will handles HTML escaping
-      return val;
+      return filterXSS(val, { stripIgnoreTag: true });
```
kristw, line 115: Could we use `dompurify` like before? 
`dompurify` seems to be more active (last updated < 1 month + 1 issue) compared to `xss` (last updated 1 year + 30 issue) with larger npm download
https://www.npmtrends.com/dompurify-vs-xss

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
```diff
@@ -203,29 +202,39 @@ export default function ReactDataTable(props: DataTableProps) {
...
+                {col.label}
+              </th>
             ))}
           </tr>
         </thead>
         <tbody>
           {data.map((record, i) => (
-            <tr key={i}>
+            // hide rows after first page makes the initial render faster (less layout computation)
+            <tr key={i} style={{ display: pageLength > 0 && i >= pageLength ? 'none' : undefined }}>
```
kristw, line 222: Even with pagination, do we still create all rows in dom?

packages/superset-ui-legacy-plugin-chart-table/src/index.ts
```diff
@@ -16,6 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+/// <reference path="../types/external.d.ts" />
```
kristw: Is this necessary?

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
ktmud: Both have decent number of followers but `js-xss` is much faster: 
https://www.measurethat.net/Benchmarks/Show/7485/0/js-xss-vs-dompurify

There is actually a noticeable speed difference when I tested in Superset with Chrome profiling tool.

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
ktmud: `jquery.datatables` either reads DOM attributes to collect data or we pass a `data` object with all the records. But the latter approach is actually slower because it also creates DOM nodes for all rows beforehand and calls "innerHTML" on every cell which we have no control of.

packages/superset-ui-legacy-plugin-chart-table/src/index.ts
ktmud: This is necessary when importing from `superset-frontend` directly. Figured it's best to leave it there since it doesn't break anything anyway.

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
kristw: I am supportive if it shows significant speed improvement. If you could note down the numbers maybe we can save that for announcement email.

BTW, is the test for `js-xss` from the link above wrong? It seems to call `sanitize`, not `filterXSS()`.
image

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
ktmud: I don't think we need the `gi` flags and captured groups. Will use `/<[^>]+>/`.

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
ktmud: My bad. Use this instead: https://www.measurethat.net/Benchmarks/Show/7485/4/dompurify-vs-js-xss

js-xss is indeed slower than [sanitize.js](https://github.com/gbirke/Sanitize.js), but still faster than dompurify, though.

kristw: sgtm. Thanks for being thorough.

kristw:
The build seems to be broken on `xss`

ktmud:
Addition information on `js-xss` vs `dompurify`, on the same 10k row sample data in Storybook, it takes ~12 secs for `dompurify` to finish rendering React DOM nodes and 3.5 secs for `js-xss`.

DOMPurify:
image

xss:
image

kristw:
* Code looks good. Some `eslint` errors breaking ci.
* Have to do something with the title.

legacy_-___plugin_-_chart_-_table___TableChartPlugin_-_Big_Table_⋅_Storybook

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
nytai: obligatory: https://stackoverflow.com/a/1732454

ktmud: Lol. We didn't really use this to extract or remove the tags, just preemptively test whether some cell text is *potentially* HTML so we can avoid using `dangerouslySetInnerHTML`.

packages/superset-ui-legacy-plugin-chart-table/src/index.ts
ktmud: Removed to fix linting error. :D

packages/superset-ui-legacy-plugin-chart-table/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@superset-ui/legacy-plugin-chart-table",
-  "version": "0.11.15",
+  "version": "0.11.5",
```
kristw: Do not modify this. This will be bumped by `lerna` when publishing. 
Also, the new version will be `0.11.16`.

packages/superset-ui-legacy-plugin-chart-table/package.json
```diff
@@ -28,12 +28,14 @@
     "access": "public"
   },
   "dependencies": {
-    "d3": "^3.5.17",
-    "datatables.net-bs": "^1.10.15",
-    "dompurify": "^2.0.6",
-    "prop-types": "^15.6.2"
+    "@superset-ui/query": "^0.12.8",
```
kristw: 1) Please move this to `peerDependencies`. 

There are several `@superset-ui` packages that use singleton so we have to ensure it has only one copy, similar to how `react` works. (Although some packages do not use singleton, I made all of them always `peerDependencies` to avoid confusion whether this should be `dependencies` or `peerDependencies`)

2) `^0.12.0` might be good enough for `@superset-ui/query`?

ktmud: Got it

packages/superset-ui-legacy-plugin-chart-table/package.json
ktmud: Oh, I meant to revert it back to 0.11.15. Previously I changed to 0.12.0.

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+// Depending on how the modules are imported, `dt` may be a CommonJS init function,
+// or the DataTable class itself. In case it is the former, we'd need to tell it
+// where is jQuery.
+if (!dt.$) {
+  dt(window, $);
+}
+
+const { PERCENT_3_POINT } = NumberFormats;
+const isPontiallyHTML = (text: string) => /<[^>]+>/.test(text);
```
kristw: * typo `isPontiallyHTML` => `isPartiallyHTML`. 
* The regex can be defined outside of the function to avoid parsing regex every function call.

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
```diff
@@ -0,0 +1,252 @@
...
+  const formatTimestamp = getTimeFormatter(tableTimestampFormat);
+  const metrics = (aggMetrics || [])
+    .concat(percentMetrics || [])
+    // actual records must be of numeric types as well
+    .filter(m => typeof data[0][m] === 'number');
+
+  // check whethere a key is a metric
+  const metricsLookup = new Set(aggMetrics);
+  const percentMetricsLookup = new Set(percentMetrics);
+  const isMetric = (key: string) => metricsLookup.has(key);
```
kristw: * perhaps rename `metricsLookup` to `metricsSet` and `percentMetricsLookup` to `percentMetricsSet`. 
* Could we call `metricsSet.has(key)` directly instead of using `isMetric` function which introduces additional function call?

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
ktmud: - Thanks for noticing the typo! I actually meant `isPotentiallyHTML`, which probably makes more sense than "partially"?
- I was hoping modern JS engines can already optimize for RegExp literals.

Perf test seems to suggest the uncached approach is actually faster:
https://www.measurethat.net/Benchmarks/Show/7486/1/regexp-literal-cache-or-not

I bet it has something to do with variable scope resolution.

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
ktmud: Sure!

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
```diff
@@ -116,13 +114,13 @@ export default function ReactDataTable(props: DataTableProps) {
...
-    if (isPercentMetric(key)) {
+    if (percentMetricsSet.has(key)) {
       // in case percent metric can specify percent format in the future
       return formatNumber(format || PERCENT_3_POINT, val as number);
     }
-    if (isMetric(key)) {
+    if (metricsSet.has(key)) {
       // default format '' will return human readable numbers (e.g. 50M, 33k)
-      return formatNumber(format || '', val as number);
+      return formatNumber(format || SMART_NUMBER, val as number);
```
kristw: no need to explicitly set `SMART_NUMBER` here. the `NumberFormatterRegistry` has ability to set a fallback if the format string is not known by the registry. The fallback is `SMART_NUMBER` by default. 

Not hardcoding it here will make it possible to override fallback at the registry.

kristw (approved):
just a few nits

packages/superset-ui-legacy-plugin-chart-table/src/ReactDataTable.tsx
ktmud: After installing all node_modules fresh, if I pass `''`, I will get a `this.get(...) is not a function` error from [NumberFormatterRegistry.ts](https://github.com/apache-superset/superset-ui/blob/9bb5c471e0e3d05da38929feb00ade7d0ecdc474/packages/superset-ui-number-format/src/NumberFormatterRegistry.ts#L42) . But if I use `formatNumber(format, val as number)`, I will get "type not supported" error, because [`formatNumber` does not accept `undefined`.](https://github.com/apache-superset/superset-ui/blob/9bb5c471e0e3d05da38929feb00ade7d0ecdc474/packages/superset-ui-number-format/src/NumberFormatterRegistrySingleton.ts#L12)

Should I just add `// @ts-ignore` above this line?

kristw: My bad. Let me fix the bug in `@superset-ui/number-format` real quick. Give 15mins.

kristw (approved):

PR #392, opened 2020-03-07
fix(legacy-table): avoid React DOM

ktmud:
`jquery.datatables` will manipulate DOMs, sometimes remove them. In case of component being reloaded with updated props, React will not be able to find those removed nodes, causing a `cannot removeChild` error.

Because of the the way to assign row keys, if table shape changes (add or remove columns), React may also have difficulty match the cached nodes via keys.

In general it's a bad idea to directly manipulate React rendered DOM nodes, so we better just let `jquery.datatables` handle everything.

In the future, once we removed `jquery.datatables`, a pure React component will not have such issues.

This PR also contains a fix to null `percent_metrics`.

cc @kristw @etr2460 

🐛 Bug Fix

### DIFF
--- a/packages/superset-ui-legacy-plugin-chart-table/test/ReactDataTable.test.tsx
+++ b/packages/superset-ui-legacy-plugin-chart-table/test/ReactDataTable.test.tsx
@@ -1,0 +1,57 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import React from 'react';
+import { mount } from 'enzyme';
+import ReactDataTable from '../src/ReactDataTable';
+import transformProps from '../src/transformProps';
+import * as testData from './test_data';
+
+describe('legacy-table', () => {
+  // Can test more prop transformation here. Not needed for now.
+  describe('transformProps', () => {});
+
+  describe('ReactDataTable', () => {
+    // Jest throw an error at `console.warn`, this disables that behavior
+    beforeAll(() => {
+      console.warn = console.log;
+    });
+    afterAll(() => {
+      console.warn = console.error;
+    });
+
+    let wrap: any; // the SuperChart wraper
+    let tree: any; // helper to dive into DOM tree
+
+    it('render basic data', () => {
+      wrap = mount(<ReactDataTable {...transformProps(testData.basic)} />);
+      tree = wrap.render(); // returns a CheerioWrapper with jQuery-like API
+      expect(tree.hasClass('superset-legacy-chart-table')).toEqual(true);
+      expect(tree.find('td[data-sort="Michael"]')).toHaveLength(1);
+    });
+
+    it('render advanced data', () => {
+      // should successfull rerender with new props
+      wrap.setProps(transformProps(testData.advanced));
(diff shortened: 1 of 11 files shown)


kristw (approved):

kristw:
Can't wait to get rid of `jquery.datatable`. sigh 
Will this change remove the performance gain we had earlier?

ktmud:
> Will this change remove the performance gain we had earlier?

I don't think so. `jquery.datatables` will still read an already constructed DOM tree. It's just an extra set `innerHTML` before the React component is mounted.

scripts/build.js
```diff
@@ -7,14 +7,18 @@ const glob = process.argv[2];
...
-  runner(p, args, { stdio: 'inherit' });
+  const runner = spawnSync;
+  const { status } = runner(p, args, { stdio: 'inherit' });
+  if (status !== 0) {
+    process.exit(status);
+  }
 };
 
 if (glob) {
+  run(`eslint --color --quiet packages/${glob}/{src,test}/**/*.{js,jsx,ts,tsx}`);
```
kristw: `nimbus eslint --color ...`

`nimbus` will create eslint config before executing. Otherwise there may not be eslint config as we don't check them in and let nimbus handle that.

ktmud: I wanted to limit to just the specified packages, but `nimbus` doesn't seem to support that for `eslint` command.

ktmud: My bad. It turns out I can just pass the file directory. Previous tried the same `--workspaces=` approach as other `nimbus` command.

packages/superset-ui-legacy-plugin-chart-table/test/test_data.ts
```diff
@@ -0,0 +1,103 @@
+/**
```
kristw: We adopt camelCase file naming.

packages/superset-ui-legacy-plugin-chart-table/test/ReactDataTable.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import React from 'react';
+import { mount } from 'enzyme';
+import ReactDataTable from '../src/ReactDataTable';
+import transformProps from '../src/transformProps';
+import * as testData from './test_data';
```
kristw: Perhaps use specific import to enable warning if we later do not use some of these.
`import { basic, advance } from ...`

packages/superset-ui-legacy-plugin-chart-table/test/ReactDataTable.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+describe('legacy-table', () => {
+  // Can test more prop transformation here. Not needed for now.
+  describe('transformProps', () => {});
+
+  describe('ReactDataTable', () => {
+    // Jest throw an error at `console.warn`, this disables that behavior
```
kristw: (optional) `nimbus` set this as default, which I also have mixed feelings. I use `jest-mock-console` in another repo, in case the error message if predictable and can be captured.

packages/superset-ui-legacy-plugin-chart-table/test/test_data.ts
ktmud: Thought it was useful to differentiate mock data with actual JS modules. But I guess it doesn't add much value anyway.

packages/superset-ui-legacy-plugin-chart-table/test/ReactDataTable.test.tsx
ktmud: Wanted to have a namespace so it's more explicit in later code, but also wanted to keep variable names simple.

packages/superset-ui-legacy-plugin-chart-table/test/ReactDataTable.test.tsx
ktmud: Removed it for now since it's not needed for this plugin anymore.

packages/superset-ui-legacy-plugin-chart-table/test/ReactDataTable.test.tsx
ktmud: Added a default export for `testData`. I think it looks better than `import *`.

In the future, maybe all plugins can reuse the same test data---similar to switching visualization type in Superset.

PR #393, opened 2020-03-08
fix(forcedirected.js): make force directed graph zoomable and pannable

### LINKED ISSUE #7246: Forced Directed graph is cut off and returns [object Object] after resize
Make sure these boxes are checked before submitting your issue - thank you!

- [ ] I have checked the superset logs for python stacktraces and included it here as text if there are any.
- [x] I have reproduced the issue with at least the latest released version of superset.
- [x] I have checked the issue tracker for the same issue and I haven't found one similar. (There is a similar issue https://github.com/apache/incubator-superset/issues/6525 that has been closed)

### Superset version
Master

### Expected results
To view the entire force directed graph and for it to continue to display after resize

### Actual results
Screen Shot 2019-04-08 at 4 10 48 PM

After resizing:
Screen Shot 2019-04-08 at 4 11 02 PM

### Steps to reproduce
1. Open sample force-directed graph Energy Force Layout
2. Resize window

IndexOutOfRange:
💔 Breaking Changes

🏆 Enhancements

📜 Documentation

🐛 Bug Fix
Fix https://github.com/apache/incubator-superset/issues/7246
Force directed graph can now be zoomed in/out and panned. 

🏠 Internal

### DIFF
--- a/packages/superset-ui-legacy-plugin-chart-force-directed/src/ForceDirected.js
+++ b/packages/superset-ui-legacy-plugin-chart-force-directed/src/ForceDirected.js
@@ -38,11 +38,18 @@ const propTypes = {
 /* Modified from http://bl.ocks.org/d3noob/5141278 */
 function ForceDirected(element, props) {
   const { data, width, height, linkLength = 200, charge = -500 } = props;
+
+  let w = window.innerWidth;
+  let h = window.innerHeight;
+  const minZoom = 0.1;
+  const maxZoom = 7;
   const div = d3.select(element);
+  const zoom = d3.behavior.zoom().scaleExtent([minZoom, maxZoom]);
   div.classed('superset-legacy-chart-force-directed', true);
 
   const links = data;
   const nodes = {};
+
   // Compute the distinct nodes from the links.
   links.forEach(link => {
     link.source =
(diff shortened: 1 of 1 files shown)


packages/superset-ui-legacy-plugin-chart-force-directed/src/ForceDirected.js
```diff
@@ -38,11 +38,18 @@ const propTypes = {
 /* Modified from http://bl.ocks.org/d3noob/5141278 */
 function ForceDirected(element, props) {
   const { data, width, height, linkLength = 200, charge = -500 } = props;
+
+  let w = window.innerWidth;
```
kristw: The chart does not always have `width` and `height` being 100% of the window `width` and `height`.

IndexOutOfRange: Thanks for the input. I commited a version where width/height are inherited.

packages/superset-ui-legacy-plugin-chart-force-directed/src/ForceDirected.js
```diff
@@ -107,15 +116,22 @@ function ForceDirected(element, props) {
     .on('tick', tick)
     .start();
 
-  div.selectAll('*').remove();
-  const svg = div
-    .append('svg')
-    .attr('width', width)
-    .attr('height', height);
+  function resize() {
+    const { innerWidth, innerHeight } = window;
```
kristw, line 120: Using `window` 's `innerWidth` and `innerHeight` makes the chart expand beyond its supposed size.

PR #397, opened 2020-03-11
fix(legacy-table): when data is empty

ktmud:
This one is pretty straightforward.

## Test Plan

A new unit test case is added.

🐛 Bug Fix

cc: @kristw @graceguo-supercat @etr2460


etr2460 (approved):
lgtm, could you add the test plan to the PR summary though?

ktmud:
> lgtm, could you add the test plan to the PR summary though?

I'll do that in the Superset PR.

etr2460:
Something as simple as:

"Added a new storybook example and ensured all the examples still render properly" is fine. It's generally a good idea to include how code was tested in the PR summary so that we can refer back to it in the future

ktmud:
Oh, I didn't add a new storybook example. Thought the unit test was enough.

PR #402, opened 2020-03-16
feat(big-number): format datetime according to granularity

ktmud:
🏆 Enhancements

This adds support for "Time Grain" in Superset for Big Number with Trendline chart. 

image

It will now render date format in tooltip according to the time granularity.

## Before

When "Time Grain" is set to "Quarter":

Snip20200315_57

## After

When "Time Grain" is set to "Quarter":

Snip20200315_56

### DIFF
--- a/packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/transformProps.js
+++ b/packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/transformProps.js
@@ -18,9 +18,44 @@
  */
 import * as color from 'd3-color';
 import { getNumberFormatter, NumberFormats } from '@superset-ui/number-format';
+import { getTimeFormatter, TimeFormats, smartDateVerboseFormatter } from '@superset-ui/time-format';
 import { renderTooltipFactory } from './BigNumber';
 
 const TIME_COLUMN = '__timestamp';
+
+function getTimeFormatterForGranularity(granularity) {
+  // Translate time granularity to d3-format
+  const MINUTE = '%Y-%m-%d %H:%M';
+  const SUNDAY_BASED_WEEK = '%Y W%U';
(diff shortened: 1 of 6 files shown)


ktmud:
Will create a [follow up PR](https://github.com/ktmud/superset-ui-plugins/pull/2/files) to introduce `alignTimeRange` after this is merged.

kristw:
Note: Can ignore the test coverage for now.

packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/transformProps.js
```diff
@@ -36,6 +71,7 @@ export default function transformProps(chartProps) {
     subheader = '',
     vizType,
   } = formData;
+  const granularity = formData.timeGrainSqla;
```
kristw, line 74: could be part of the destructuring above

packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/transformProps.js
```diff
@@ -18,10 +18,45 @@
  */
 import * as color from 'd3-color';
 import { getNumberFormatter, NumberFormats } from '@superset-ui/number-format';
+import { getTimeFormatter, TimeFormats, smartDateVerboseFormatter } from '@superset-ui/time-format';
 import { renderTooltipFactory } from './BigNumber';
 
 const TIME_COLUMN = '__timestamp';
 
+function getTimeFormatterForGranularity(granularity) {
```
kristw, line 26: This seems useful to extract into its own file, in case we move this to `@superset-ui/time-format` later

ktmud:
Merge it for now. No need to publish. Will address the comments in a future PR.

PR #403, opened 2020-03-17
feat(big-number): allow fallback to last available value and fix time range for trend lines

ktmud:
🏆 Enhancements

The Big Number chart has two problems related to the handling of null values:

1. **Misleading time range**: When there is no data at the beginning or ending periods of the filtered time range, the backend returns no data records for those periods, therefore the trend line will only start rendering from the datetime where there is data. This often leads to confusion or misinterpretations, especially then in the dashboards that have multiple trend line charts filtered to the same time range. They could all be showing similar smooth lines, yet representing very different time ranges.
2. **NULL value at the last record**: When the last record returns NULL value for the metric, the Big Number text will render as "No data", but the trend line still renders. Some users have requested to fallback displaying the Big Number with the last available dates.

This PR adds option `timeRangeFixed` to fix the first problem and adds a "fallback to last available value" behavior to address the second problem.

The `timeRangedFixed` option is disabled by default. When enabled, the empty starting and ending periods will take space on the x-axis, but there will be no connected lines, dots, or tooltips, just as if the metric values are NULL.

In case the last record has a NULL value for the metric, we will go back to the last available value and display it in grayed out style (see screen shots below).

Superset backend needs to be updated to return `query.from_dttm` and `query.to_dttm`. Frontend needs to be updated to add two checkbox controls for these options ( https://github.com/apache/incubator-superset/pull/9341). 

This PR also rewrote the Big Number chart plugin with Typescript.

#### Before

Snip20200316_62

#### After

Snip20200316_59

image

Above chart has a time range between 2001 to 2014, yet only since 2006 it starts to have data. When `timeRangeFixed` is enabled, it will render from the beginning of the parsed filter time range (2001), instead of when it starts to actually have data (2006).

##### Before 

image

##### After

Snip20200325_15

I also increased the default font sizes a little bit.

##### After

toggle-fixed-time-range

### Test Plan

Go to storybook and make sure all examples for Big Number chart work.

There is also a new unit test to ensure `timeRangeUseFallback` works.

### Related

https://github.com/apache-superset/superset-ui-plugins/pull/402
https://github.com/ktmud/superset-ui-plugins/pull/2

## Reviewer
cc @kristw @etr2460


packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/BigNumber.tsx
```diff
@@ -185,21 +191,33 @@ class BigNumberVis extends React.PureComponent {
...
       mainColor,
       subheader,
       renderTooltip,
       startYAxisAtZero,
+      fromDatetime,
+      toDatetime,
+      useFixedTimeRange,
     } = this.props;
 
+    // Apply a fixed tange range if a time range is specified.
```
kristw: typo `tange` => `time`

packages/superset-ui-legacy-preset-chart-big-number/package.json
```diff
@@ -42,5 +41,8 @@
     "@superset-ui/time-format": "^0.12.0",
     "@superset-ui/translation": "^0.12.0",
     "react": "^15 || ^16"
+  },
+  "devDependencies": {
+    "@types/shortid": "0.0.29"
```
kristw: Let's put this in `dependencies`
https://stackoverflow.com/questions/45176661/how-do-i-decide-whether-types-goes-into-dependencies-or-devdependencies

packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/BigNumber.tsx
```diff
@@ -41,65 +43,69 @@ const PROPORTION = {
...
+    subheader: '',
+    subheaderFontSize: PROPORTION.SUBHEADER,
+    toDatetime: null,
+    trendLineData: null,
+    useFixedTimeRange: false,
+  };
+
+  constructor(props: BigNumberVisProps) {
     super(props);
     this.gradientId = shortid.generate();
```
kristw: could do 

```ts
private gradientId: string = shortid.generate();
```

and omit the constructor

kristw (commented):
There are `eslint` errors.

```
/home/travis/build/apache-superset/superset-ui-plugins/packages/superset-ui-legacy-preset-chart-big-number/src/index.js
  1:49  error  Missing file extension "ts" for "./BigNumber/index"       import/extensions
  2:54  error  Missing file extension "ts" for "./BigNumberTotal/index"  import/extensions
/home/travis/build/apache-superset/superset-ui-plugins/packages/superset-ui-legacy-preset-chart-big-number/src/preset.js
  20:34  error  Missing file extension "ts" for "./BigNumber"       import/extensions
  21:39  error  Missing file extension "ts" for "./BigNumberTotal"  import/extensions
```

Don't add the `ts` extension like it suggests. Just apply `eslint-disable` because once these file are transpiled into `js` in `lib` and `esm`, having the `.ts` extension in `import` statement will make it fail to find the file.

ktmud:
> Don't add the `ts` extension like it suggests. Just apply `eslint-disable` because once these file are transpiled into `js` in `lib` and `esm`, having the `.ts` extension in `import` statement will make it fail to find the file.

I moved all files to `ts` instead and made Storybook read `index.ts` as well: https://github.com/apache-superset/superset-ui-plugins/pull/403/files#diff-55dda99e8a651cb4fdb5eb443cb1f8b5R50

packages/superset-ui-legacy-preset-chart-big-number/package.json
```diff
@@ -29,11 +29,11 @@
   },
   "dependencies": {
     "@data-ui/xy-chart": "^0.0.84",
-    "d3-color": "^1.2.3",
-    "prop-types": "^15.6.2",
+    "@types/shortid": "0.0.29",
     "shortid": "^2.2.14"
   },
   "peerDependencies": {
+    "d3-color": "^1.2.3",
```
kristw: Let's keep this one as `dependencies`.

ktmud: Any reason why would it be better? I thought it's preferable to have it in `peerDependencies` as long as it is in the `dependencies` of `superset-frontend`.

kristw: Usually `peerDependencies` should be use when only one instance of that lib should exist in the entire app. This could be due to the use of singleton or because the dependency library is very heavy. The downside of `peerDependencies` is it put more responsibility for the app developers who are consuming the library.

* Need to add that library to the app-level `dependencies`. 
* If every lib lists everything as `peerDependencies`, the app will have a long list of `dependencies` that many of them are not used directly in the app code.
* If this lib stop using `d3-color`, app developers need to remove it from their app's `dependencies` or it will become redundant dependencies.
* If the app needs `d3-color` version that is not compatible with this library, then there is a conflict.

So if the library is small and does not have singleton requirement, we should list as `dependencies` and rely on `npm` or `yarn` to handle that. This will encapsulate the dependencies and avoid the additional task for app developers to manage `d3-color` as its `dependencies`.

ktmud: Understood. And this totally makes sense. I'm wondering whether we can issue an "official" list of `peerDependencies` supported by Superset that plugin developers can trust will always be available and stable. I'd imagine `d3` packages to be one of them.

packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/transformProps.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  let bigNumber = null;
+  let trendLineData = null;
+  let percentChange = 0;
+
+  bigNumber = data.length === 0 ? null : data[0][metricName];
+
+  if (data.length > 0) {
+    const sortedData = [...data]
```
kristw: Could use `data.map` directly instead of `[...data].map` because `.map` already create a copy.

packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/transformProps.ts
```diff
@@ -0,0 +1,132 @@
...
+  const supportAndShowTrendLine = supportTrendLine && showTrendLine;
+  let formattedSubheader = subheader;
+
+  let mainColor;
+  if (colorPicker) {
+    const { r, g, b } = colorPicker;
+    mainColor = color.rgb(r, g, b).hex();
+  }
+
+  let bigNumber = null;
```
kristw: Can merge with line 61

```ts
let bigNumber = data.length === 0 ? null : data[0][metricName];
```

packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/transformProps.ts
```diff
@@ -0,0 +1,132 @@
...
+    headerFontSize,
+    metric,
+    showTrendLine,
+    startYAxisAtZero,
+    subheader = '',
+    subheaderFontSize,
+    timeGrainSqla: granularity,
+    vizType,
+    timeRangeFixed = false,
+    timeRangeUseFallback = false,
```
kristw: Perhaps rename this field and change from `boolean` to `bigNumberMode: 'last' | 'lastAvailable'`.
We could extend this to support additional modes: `'average'`, `'sum'` in the future.

packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/transformProps.ts
```diff
@@ -0,0 +1,132 @@
...
+    compareSuffix = '',
+    headerFontSize,
+    metric,
+    showTrendLine,
+    startYAxisAtZero,
+    subheader = '',
+    subheaderFontSize,
+    timeGrainSqla: granularity,
+    vizType,
+    timeRangeFixed = false,
```
kristw, line 47: Is this the old `useFixedTimeRange` field?

packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/BigNumber.tsx
```diff
@@ -185,32 +191,65 @@ class BigNumberVis extends React.PureComponent {
...
+      fromDatetime,
+      timeRangeFixed,
     } = this.props;
 
+    // Apply a fixed X range if a time range is specified.
+    //
+    // XYChart checks the existence of `domain` property decide whether to apply
+    // a domain or not, so it must not be `null` or `undefined`
+    const xScale: { type: string; domain?: number[] } = { type: 'timeUtc' };
+    const tooltipData = trendLineData?.sort(datum => datum.x);
```
kristw: * This `.sort()` will mutate the original `props`, which violates `react` 's one-way data flow concept. Should make a copy first `trendlineData?.concat().sort(...)`
* It seems like `trendLineData` should already have been sorted in `transformProps`?

packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/BigNumber.tsx
```diff
@@ -185,32 +191,65 @@ class BigNumberVis extends React.PureComponent {
...
+    if (tooltipData && timeRangeFixed && fromDatetime) {
+      let { toDatetime } = this.props;
```
kristw:
```ts
const toDateTime = this.props.toDateTime ?? Date.now();
```

packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/BigNumber.tsx
```diff
@@ -185,32 +191,65 @@ class BigNumberVis extends React.PureComponent {
...
+      if (tooltipData[tooltipData.length - 1].x < toDatetime) {
+        tooltipData.push({
+          x: toDatetime,
+          y: null,
+        });
+      }
+      xScale.domain = [fromDatetime, toDatetime];
+    }
+    // if can't find any non-null values, no point rendering the trendline
+    if (!tooltipData?.some(d => d.y !== null)) {
```
kristw: Can move this `if` block above the block before it.

kristw (commented):
Agree with the direction of the solution.

packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/transformProps.ts
ktmud: Yes

packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/transformProps.ts
ktmud: In that case, should we use a `SelectControl` in Superset, too? I'm a little hesitant to add this for now. Allowing `sum`, `average` or any other computation mode significantly increases the complexity of how the product can be used and may cause confusion among end users. Even if the chart builders know what they are doing, the readers might not. Chart builders will need to add a note to the dashboard explaining what the big number is, or we'd have to do that somewhere in the chart.

kristw: I am just afraid to add this `timeRangeUseFallback` field and soon have to write `db:migration` for it. 
At least, could rename it to `useLastAvailableDataPoint` to be more specific.

For more context: 

The current `BigNumber` and `BigNumberTotal` are behaving differently. One show last value and one show sum. 
There was a feature request to clarify this at one point. 

The returned data from the legacy API are also different.

`BigNumber`: array of value over time

```
[
  {
    __timestamp: 1388534400000.0,
    sum__SP_POP_TOTL: 7237260256.0,
  },
  {
    __timestamp: 1356998400000.0,
    sum__SP_POP_TOTL: 7151135481.0,
  },
  {
    __timestamp: 1325376000000.0,
    sum__SP_POP_TOTL: 7066007165.0,
  },
  {
    __timestamp: 1293840000000.0,
    sum__SP_POP_TOTL: 6984252419.0,
  }
]
```

`BigNumberTotal`: sum as single entry in the array

```
[
  {
    sum__num: 32546308,
  },
]
```

ktmud: Hmm. Why does this sound intuitive and surprising at the same time... 😂 Still, I worry it will be probably even more confusing if Big Number with Trend Line can display sums, too.

packages/superset-ui-legacy-preset-chart-big-number/package.json
kristw: > "official" list of peerDependencies supported by Superset that plugin developers can trust will always be available and stable. I'd imagine d3 packages to be one of them.

Each plugin may be developed on current version of d3 but then later on if d3 upgrades over time, but some plugins did not migrate, some did, it will cause problem what version Superset should say official, because everybody tries to point to the same d3. At this point, it is not clear what value we actually gain from having d3 as `peerDependencies`. It is better to use `dependencies`. 

Also, these plugins can also be used in another app that is not Superset.

packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/transformProps.ts
kristw: Another alternative solution is to fallback to last available data point by default, but add warning to indicate that the last data point is missing.

kristw: options:
1. current solution but rename variable from `timeRangeUseFallback` to `useLastAvailableDataPoint` (or something that is more descriptive).
2. fallback to last available data point by default, but add warning to indicate that the last data point is missing. (maybe your current solution 1. is already better, or can circle back with @etr2460)
3. changes from boolean to string. 

I'm good with any one them.

ktmud: I like your alternative solution. Maybe display the Big Number grayed out and add a tooltip explaining this is the last available value. It saves the trouble for future migration and makes sure chart readers understand what the value is, too.

packages/superset-ui-legacy-preset-chart-big-number/package.json
ktmud: I'd say packing multi versions of d3 to the client side is still something we should actively try to avoid. Maybe incorporate something like [InspectPack](https://github.com/formidablelabs/inspectpack/) would help.

packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/transformProps.ts
ktmud: image

Come up with this: fallback to the last available value by default + add some hover tooltip to the fallback values.

packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/BigNumber.tsx
```diff
@@ -0,0 +1,309 @@
...
+  timeRangeFixed: boolean;
+  trendLineData?: TimeSeriesDatum[];
+  mainColor: string;
+};
+
+class BigNumberVis extends React.PureComponent<BigNumberVisProps, {}> {
+  private gradientId: string = shortid.generate();
+
+  static defaultProps = {
+    bigNumberFallback: null,
```
kristw: Do not need to specify `null` as default props and use default behavior `undefined`. 
`null` will also do not match the type `bigNumberFallback?: TimeSeriesDatum;` which is `TimeSeriesDatum | undefined`.

Same with `fromDatetime`, `toDatetime`, and `trendLineData`

packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/BigNumber.tsx
```diff
@@ -0,0 +1,309 @@
...
+
+type BigNumberVisProps = {
+  className?: string;
+  width: number;
+  height: number;
+  bigNumber: number;
+  bigNumberFallback?: TimeSeriesDatum;
+  formatNumber: NumberFormatter;
+  formatTime: TimeFormatter;
+  fromDatetime: number;
```
kristw: Should this be `fromDatetime?` since the field is optional ? 
Same with `toDatetime` and `trendLineData`

packages/superset-ui-legacy-preset-chart-big-number/src/BigNumber/BigNumber.tsx
```diff
@@ -0,0 +1,309 @@
...
+import { XYChart, AreaSeries, CrossHair, LinearGradient } from '@data-ui/xy-chart';
+import { BRAND_COLOR } from '@superset-ui/color';
+import { computeMaxFontSize } from '@superset-ui/dimension';
+import NumberFormatter from '@superset-ui/number-format/lib/NumberFormatter';
+import { smartDateVerboseFormatter } from '@superset-ui/time-format';
+import TimeFormatter from '@superset-ui/time-format/lib/TimeFormatter';
+
+import './BigNumber.css';
+
+const defaultNumberFormatter = getNumberFormatter(undefined);
```
kristw: Is it possible to omit `undefined`?

ktmud: Just wanted to be explicit. `getNumberFormatter` sounds like something that needs a parameter.

kristw (approved):

kristw:
Very nice. Could we move the big number up a bit in case of fallback so they align?
That's last thing and I'll merge.
Pasted_Image_3_27_20__4_50_PM

package.json
```diff
@@ -42,12 +42,12 @@
   ],
   "license": "Apache-2.0",
   "devDependencies": {
-    "@airbnb/config-babel": "^2.1.3",
-    "@airbnb/config-eslint": "^2.1.3",
-    "@airbnb/config-jest": "^2.1.3",
-    "@airbnb/config-prettier": "^2.0.4",
-    "@airbnb/config-typescript": "^2.1.2",
-    "@airbnb/nimbus": "^2.1.3",
+    "@airbnb/config-babel": "2.2.4",
```
kristw: Why pin to specific version without `^`?

packages/superset-ui-legacy-preset-chart-big-number/package.json
```diff
@@ -29,8 +29,9 @@
   },
   "dependencies": {
     "@data-ui/xy-chart": "^0.0.84",
+    "@types/d3-color": "^1.2.2",
+    "@types/shortid": "0.0.29",
```
kristw: same here

package.json
ktmud: Was having trouble building. Couldn't figure out why. Just don't want to deal with the unpredictability. Sometimes even with the implied "not a breaking change" in a minor version upgrade, things might still break.

packages/superset-ui-legacy-preset-chart-big-number/package.json
kristw: This should use the `^`

package.json
kristw: This repo already check in `yarn.lock` so if you did not modify these package version, it should be the same version that was working in `master`. Could you try reseting `yarn.lock` and `package.json` to `master`'s version while keeping the changes you made to `BigNumber`'s `package.json` and `yarn install` again?

ktmud: Weird, it works... I'm pretty sure I tried that yesterday... Maybe it's because I inadvertently pined `@types/shortid`.  

I think it's probably not a bad idea to pin `devDependencies` as they are unlikely to cause dependency hell and the added stability is valuable. Renovate actually [pin the dependencies for you](https://github.com/ktmud/incubator-superset/pull/2) once you set it up.

